PR #1, opened 2022-05-30
CTCP-419: create base for functioning router

derekspalmer:


RobertLaverick (dismissed):

project/CodeCoverageSettings.scala
```diff
@@ -16,7 +16,7 @@ object CodeCoverageSettings {
 
   val settings: Seq[Setting[_]] = Seq(
     ScoverageKeys.coverageExcludedPackages := excludedPackages.mkString(";"),
-    ScoverageKeys.coverageMinimumStmtTotal := 100,
+    ScoverageKeys.coverageMinimumStmtTotal := 60,  // TODO: increase as we progress.
```
drnaylor-oc, line 19: If you turn the `appName` in `AppConfig` into a `lazy val`, it then doesn't get counted for coverage, meaning we should be able to bump this back up to 90.

drnaylor-oc: Try it! Change `AppConfig.appName` to be a `lazy val` and run `./run_all_tests.sh` locally - you should get 100% because the only code left to test is the hello world service. If you look in the coverage report, you'll see lazy vals don't come back red or green - i.e., not counted for coverage.

drnaylor-oc (approved):

PR #6, opened 2022-06-14
Ctcp 514 service

derekspalmer:


app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  //TODO:  Not quite sure what we return here
+  private def post(
+    office: Future[ParseResult[OfficeOfDeparture]]
+  )(implicit messageSender: MessageSender, body: Source[ByteString, _], hc: HeaderCarrier): Future[Either[ParseError, Int]] = {
+    import concurrent.ExecutionContext.Implicits.global
+    office.flatMap {
+      case Right(office) =>
+        postToEis(office).map {
+          case Right(_)    => Right(play.api.http.Status.ACCEPTED)
```
liandra-giacomine, line 107: Should we return unit instead of the status code? If it succeeds the status code returned will always be ACCEPTED, so that's something that could be derived in the controller if it gets a Right

derekspalmer: Agreed - will return Unit.

app/uk/gov/hmrc/transitmovementsrouter/services/XmlParser.scala
```diff
@@ -0,0 +1,54 @@
...
+    .single("messageSender")
+
+  val officeOfDepartureExtractor: Flow[ParseEvent, ParseResult[OfficeOfDeparture], NotUsed] = XmlParsing
+    .subtree("CC015C" :: "CustomsOfficeOfDeparture" :: "referenceNumber" :: Nil)
+    .collect {
+      case element if element.getTextContent.nonEmpty => OfficeOfDeparture(element.getTextContent)
+    }
+    .single("referenceNumber")
+
+  def messageSenderWriter(messageId: MessageId): Flow[ParseEvent, ParseEvent, NotUsed] = Flow[ParseEvent]
```
drnaylor-oc, line 41: It's not the message ID we want to set here, but the `MessageSender`

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -0,0 +1,124 @@
...
+
+        val xmlParsing: FlowShape[ByteString, ParseEvent]                            = builder.add(XmlParsing.parser)
+        val officeOfDeparture: FlowShape[ParseEvent, ParseResult[OfficeOfDeparture]] = builder.add(XmlParser.officeOfDepartureExtractor)
+        xmlParsing ~> officeOfDeparture ~> officeShape
+
+        SinkShape(xmlParsing.in)
+    }
+  )
+
+  def buildMessage(messageId: MessageId): Flow[ByteString, ByteString, NotUsed] =
```
drnaylor-oc, line 70: Should be the `MessageSender`, not the `MessageId`.

test/uk/gov/hmrc/transitmovementsrouter/services/XmlParserSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+        _.right.get mustBe MessageSenderId("GB1234")
+      }
+    }
+
+    "when provided with a missing messageSender node" in {
+      val stream       = createParsingEventStream(cc015cNoMessageSenderNode)
+      val parsedResult = stream.via(XmlParser.messageSenderExtractor).runWith(Sink.head)
+
+      whenReady(parsedResult) {
+        _.isLeft
```
drnaylor-oc, line 50: We should check that the correct error is returned, not just that one is returned.

derekspalmer: Agree. Will change.

app/uk/gov/hmrc/transitmovementsrouter/services/XmlParser.scala
derekspalmer: Agree.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
derekspalmer: Agree.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -0,0 +1,124 @@
...
+  )(implicit hc: HeaderCarrier): EitherT[Future, RoutingError, Unit] = {
+
+    val officeOfDeparture       = payload.toMat(officeOfDepartureSink)(Keep.right).run()
+    implicit val messageSender  = MessageSender(movementId, messageId)
+    implicit val updatedPayload = payload.via(buildMessage(messageSender))
+
+    EitherT(post(officeOfDeparture))
+  }
+
+  //TODO:  Not quite sure what we return here
```
crystalmark, line 99: We should be returning a 202 Accepted (for now) which is what we are expecting from EIS[-stub]

derekspalmer: Yep.

derekspalmer: Forgot to remove the comment!

drnaylor-oc: Yes - though in this layer we're just sending back unit at the moment because we have nothing to report other than a success. The controller will return a 202 upon seeing a unit.

In the future, I suspect this will be an `Option[Source[ByteString, _]]` if we have synch responses and can harness the power of streaming responses.

crystalmark (approved):

PR #10, opened 2022-09-12
CTCP-980

RobertLaverick:
CTCP-980

Added incoming endpoint to allow EIS messages to be received and forward to the transit-movements service


conf/application.conf
```diff
@@ -159,7 +159,17 @@ microservice {
...
+
+    transit-movements {
+      protocol = "http"
+      host = "localhost"
+      port = 9520
+    }
   }
 
+
+messageSizeLimit = 500000
```
drnaylor-oc, line 171: Move this under the brace below.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessageController.scala
```diff
@@ -58,4 +66,18 @@ class MessagesController @Inject() (
       )
   }
 
+  def incoming(ids: (MovementId, MessageId)): Action[Source[ByteString, _]] =
+    (DefaultActionBuilder.apply(cc.parsers.anyContent) andThen messageSize()).async(streamFromFile) {
```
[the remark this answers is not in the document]
drnaylor-oc: More so with arrivals, that could in theory happen - it'll all come out in the wash with the large message discovery!

app/uk/gov/hmrc/transitmovementsrouter/services/StreamingMessageTrimmer.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def trim(source: Source[ByteString, _], messageType: MessageType): Source[ByteString, _]
+}
+
+class StreamingMessageTrimmerImpl @Inject() extends StreamingMessageTrimmer {
+
+  override def trim(source: Source[ByteString, _], messageType: MessageType): Source[ByteString, _] =
+    source
+      .via(XmlParsing.parser)
+      .via(XmlParsing.subslice("TraderChannelResponse" :: messageType.rootNode :: Nil))
```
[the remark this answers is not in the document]
RobertLaverick: So this has changed, since this comment (it sliced wrong, whoops) - but essentially the TraderChannelResponse tag is entirely andEIS wrapper, NCTS doesn't include it, we don't care to keep it and we don't intend to forward it along to the third parties in any way, so we remove it before we forward it to be persisted or pushed to the third parties.

drnaylor-oc:
I got a few warnings when compiling this - a few unused imports that need cleaning up but a couple of more substantive ones.

A match statement needs updating to be exhaustive:

```
[warn] /Users/danielnaylor/HMRC-CTC/Code/transit-movements-router/app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala:78:41: match may not be exhaustive.
[warn] It would fail on the following inputs: BadDateTime(_, _), NoElementFound(_), TooManyElementsFound(_)
[warn]     val message: String = response.left.get match {
```

and in the tests:

```
[warn] /Users/danielnaylor/HMRC-CTC/Code/transit-movements-router/test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala:99:52: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: ReverseMessagesController.incoming(ids: (uk.gov.hmrc.transitmovementsrouter.models.MovementId, uk.gov.hmrc.transitmovementsrouter.models.MessageId)): play.api.mvc.Call
[warn]   given arguments: movementId, messageId
[warn]  after adaptation: ReverseMessagesController.incoming((movementId, messageId): (uk.gov.hmrc.transitmovementsrouter.models.MovementId, uk.gov.hmrc.transitmovementsrouter.models.MessageId))
[warn]   val incoming = routes.MessagesController.incoming(movementId, messageId).url
```

It is what we want in this case, but we should adapt the call to be a tuple.

test/uk/gov/hmrc/transitmovementsrouter/services/StreamingMessageTrimmerSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+import scala.util.Failure
+import scala.util.Success
+
+class StreamingMessageTrimmerSpec extends AnyFreeSpec with Matchers with ScalaFutures with TestActorSystem {
+  "trim flow should" - {
+
+    val sut = new StreamingMessageTrimmerImpl
+
+    "successfully trim with valid xml" in {
+      val input = "<TraderChannelResponse><test>text</test></TraderChannelResponse>"
```
liandra-giacomine, line 38: Is TraderChannelResponse going to be the first element in the message? I don't think the subslice will work if it's embedded, in that case you'd need to provide the path from the root node

drnaylor-oc: It will be - EIS wrap the XML and this should be the root element.

drnaylor-oc (approved):

PR #12, opened 2022-09-15
CTCP1012 Add x-message-id header in response

ardwalker:


liandra-giacomine (dismissed):

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -77,8 +78,9 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
         .withBody(source)
         .execute[Either[UpstreamErrorResponse, HttpResponse]]
         .map {
-          case Right(_) =>
-            Right(())
+          case Right(res) =>
+            val m = Json.toJson(res.body) \ "messageId"
+            Right(MessageId(m.head.toString))
```
RobertLaverick, line 83: You wouldn't need to do this if there is a format or reads defined for MessageId

`  implicit lazy val messageIdFormat: Format[MessageId] = Json.valueFormat[MessageId]
` for example.

RobertLaverick: Also you need to update the persistence connector spec

RobertLaverick (changes requested):
See inline comments about
1) Using a format or reads for the messageId
2) Adding the test changes to the persistence connector to confirm it's returning when not mocked.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -77,8 +78,9 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
         .withBody(source)
         .execute[Either[UpstreamErrorResponse, HttpResponse]]
         .map {
-          case Right(_) =>
-            Right(())
+          case Right(res) =>
+            val m = Json.toJson(res.body) \ "messageId"
```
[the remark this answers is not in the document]
drnaylor-oc: if we don't gett a message ID, it's our problem, so would be a 5xx (but that would be a job for the controller, not this, which is a response to our request.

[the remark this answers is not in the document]
drnaylor-oc: I wonder, however, if this should get a domain object/DTO and we should return that, rather than just pulling the message ID out here, especially if we need to return more information later (thinking PPNS)

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -77,8 +78,11 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
         .withBody(source)
         .execute[Either[UpstreamErrorResponse, HttpResponse]]
         .map {
-          case Right(_) =>
-            Right(())
+          case Right(res) =>
+            Json
+              .fromJson[MessageId](Json.parse(res.body))
```
drnaylor-oc, line 83: ...wait, does this work?

The body you're expecting is 

```
{
    "messageId": "ksjdfhgkjhsdgb"
}
```

`MessageId` is just a value, that would just try to parse a `JsString`...

drnaylor-oc: Oh, I see what you've done later on...

app/uk/gov/hmrc/transitmovementsrouter/models/MessageId.scala
```diff
@@ -16,4 +16,12 @@
...
 package uk.gov.hmrc.transitmovementsrouter.models
 
+import play.api.libs.json.Reads
+import play.api.libs.json.__
+
 case class MessageId(value: String) extends AnyVal
+
+object MessageId {
+
+  implicit val messageIdReads: Reads[MessageId] = (__ \ "messageId").read[String].map(MessageId(_))
```
drnaylor-oc, line 26: I suspect this'll get confusing if we try to either:

* Expand the response object
* Try to read the MessageId from elsewhere

I think I'd prefer to see a separate response object that contains the message ID, perhaps

```scala
case class PersistenceResponse(messageId: MessageId)
```

then we can just use

```scala
implicit val persistenceResponseReads = Json.reads[PersistenceResponse]
```

or whatever it is, return that from the connector, and get the message ID out where we need it.

liandra-giacomine: Good point Dan, made the update

RobertLaverick (approved):

drnaylor-oc (approved):

PR #14, opened 2022-09-22
Send messages from trader via transit-movements-router to EIS/ERMIS

liandra-giacomine:


test/uk/gov/hmrc/transitmovementsrouter/services/XmlParserSpec.scala
```diff
@@ -62,62 +61,87 @@ class XmlParserSpec extends AnyFreeSpec with TestActorSystem with Matchers {
...
-    }
   }
 
-  "MessageSenderElement parser should add the messageSender node with the movement Id" in new Setup {
+  trait Setup {
 
-    val stream = createStream(cc015cNoMessageSenderNode)
+    def messageOfficeOfDepartureGb(messageTypeNode: String): NodeSeq = {
+      val strMessage =
+        s"<$messageTypeNode><preparationDateAndTime>2022-03-28T09:15:12</preparationDateAndTime><CustomsOfficeOfDeparture><referenceNumber>GB6789</referenceNumber></CustomsOfficeOfDeparture></$messageTypeNode>"
```
drnaylor-oc, line 118: These nodes aren't technically correct with regards to what you'll receive. I think your actual XML parsing code is right though -- it's all to do with the namespaces.

The XML root node should look something like:

```xml
<ncts:CC014C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec"></ncts:CC014C>
```

We should update all tests to use this.

test/uk/gov/hmrc/transitmovementsrouter/services/RoutingServiceSpec.scala
```diff
@@ -38,38 +38,58 @@ import scala.xml.NodeSeq
...
-      }
+  "Submitting a declaration request" - new Setup {
+
+    MessageType.departureRequestValues.foreach {
+      messageType =>
+        s"${messageType.code} should generate a valid departure office and updated payload for a GB payload" in {
+          val serviceUnderTest = new RoutingServiceImpl(mockMessageConnectorProvider)
+          val payload          = createStream(messageOfficeOfDepartureGB(messageType.rootNode))
+          val response = serviceUnderTest.submitMessage(
+            MovementType("Departure"),
```
drnaylor-oc, line 49: I've come to the realisation that movement type is useless here. If we were really using it, we'd have to make this a sealed trait with two objects, because we have Departure here, and departures elsewhere -- yet we're using the X-Message-Type header for what this was meant to do, agliee stylee and all that.

I've raised CTCP-1337 to potentially remove this -- but that's not for this PR.

drnaylor-oc:
I've just updated the tests to use property based testing and to fix a comment that had been missed. The tests all pass, and I tested using [this branch of the API layer](https://github.com/hmrc/common-transit-convention-traders/commit/5d19d7e3d815d7e4c8e00b0a9faf0e4c434cf10c) to check that it works with the header, and I'm pleased to say it does.

This either needs to await the XML arrivals endpoint PR, or for the linked branch to be PRd and merged -- one for standup tomorrow.

derekspalmer (commented):
If we're not checking for Unit do we even need to do this?

drnaylor-oc:
> If we're not checking for Unit do we even need to do this?

Yes, we're still checking that the future was succesful and not a failure, an exception will be thrown by `whenReady` if it's a failure. (I took out checking for Unit because it was causing compiler warnings.)

PR #15, opened 2022-10-07
Ctcp 1291

derekspalmer:


app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -91,15 +107,35 @@ class RoutingServiceImpl @Inject() (messageConnectorProvider: EISConnectorProvid
...
+        val message = s"An unexpected error occurred - got a ${msgTypeError.code}"
+        logger.error(message)
+        Left(RoutingError.Unexpected(message, None))
+    }
+
+    val materializedResult: Either[RoutingError, Future[ParseResult[CustomsOffice]]] = sink
+      .map(
+        payload
+          .toMat(_)(Keep.right)
+          .run()
```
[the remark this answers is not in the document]
derekspalmer: I think .map(payload.runWith) is fine actually.

app/uk/gov/hmrc/transitmovementsrouter/models/CustomsOffice.scala
```diff
@@ -16,7 +16,7 @@
 
 package uk.gov.hmrc.transitmovementsrouter.models
 
-case class OfficeOfDeparture(value: String) extends AnyVal {
+case class CustomsOffice(value: String) extends AnyVal {
   def isGB: Boolean = value.startsWith("GB")
   def isXi: Boolean = !isGB
```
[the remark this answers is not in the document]
derekspalmer: Not sure.  May be we need to know the requirement from a business perspective.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -46,28 +46,44 @@ trait RoutingService {
...
+        val xmlParsing: FlowShape[ByteString, ParseEvent] = builder.add(XmlParsing.parser)
+        val customsOffice: FlowShape[ParseEvent, ParseResult[CustomsOffice]] =
+          builder.add(XmlParser.customsOfficeExtractor(messageType, "CustomsOfficeOfDeparture"))
+        xmlParsing ~> customsOffice ~> officeShape
 
         SinkShape(xmlParsing.in)
     }
   )
 
+  private def officeOfDestinationActualSink(messageType: MessageType): Sink[ByteString, Future[ParseResult[CustomsOffice]]] =
```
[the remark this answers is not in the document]
derekspalmer: Yes, agree. We determine the node once we know whether it's an arrival request type or a departure request type.

app/uk/gov/hmrc/transitmovementsrouter/services/XmlParser.scala
```diff
@@ -31,10 +31,13 @@ object XmlParser extends XmlParsingServiceHelpers {
     }
     .single("messageSender")
 
-  def officeOfDepartureExtractor(messageType: MessageType): Flow[ParseEvent, ParseResult[OfficeOfDeparture], NotUsed] = XmlParsing
-    .subtree(messageType.rootNode :: "CustomsOfficeOfDeparture" :: "referenceNumber" :: Nil)
+  def customsOfficeExtractor(
+    messageType: MessageType,
+    officeNode: String
+  ): Flow[ParseEvent, ParseResult[CustomsOffice], NotUsed] = XmlParsing
+    .subtree(messageType.rootNode :: officeNode :: "referenceNumber" :: Nil)
```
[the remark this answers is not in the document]
derekspalmer: Yes. I like that.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -91,15 +107,35 @@ class RoutingServiceImpl @Inject() (messageConnectorProvider: EISConnectorProvid
     payload: Source[ByteString, _]
   )(implicit hc: HeaderCarrier): EitherT[Future, RoutingError, Unit] = {
 
-    val officeOfDeparture       = payload.toMat(officeOfDepartureSink(messageType))(Keep.right).run()
+    val sink: Either[RoutingError, Sink[ByteString, Future[ParseResult[CustomsOffice]]]] = messageType match {
```
liandra-giacomine, line 110: Can we update MovementType to be a sealed abstract class as we have defined done: https://github.com/hmrc/transit-movements/blob/2441c9e9b2f37b16d96fd7417f2ea1a60386a80e/app/uk/gov/hmrc/transitmovements/models/MovementType.scala

It should only ever equal departure or arrival

derekspalmer: I agree but we should create a new ticket for this, otherwise we'll get ticket scope creep.

app/uk/gov/hmrc/transitmovementsrouter/models/CustomsOffice.scala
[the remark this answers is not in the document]
derekspalmer: Agree. I'll change to check explicitly for Xi.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
drnaylor-oc: Are we actaully using `MovementType` anyway? This is `messageType` here, and I couldn't see any usage of `MovementType`.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -91,15 +107,35 @@ class RoutingServiceImpl @Inject() (messageConnectorProvider: EISConnectorProvid
...
+      case _: ArrivalRequestMessageType   => Right(officeOfDestinationActualSink(messageType))
+      case _: DepartureRequestMessageType => Right(officeOfDepartureSink(messageType))
+      case msgTypeError =>
```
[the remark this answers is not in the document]
derekspalmer: Refactor to only accept only a request message types.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
derekspalmer: We don't use messageType for this. The point regarding restricting the values is valid, but may be for another ticket.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
```diff
@@ -64,19 +66,29 @@ object MessageType {
   // *******************
 
   /** E_DEC_AMD (IE013) */
-  case object DeclarationAmendment extends DepartureRequestMessageType("IE013", "CC013C")
+  case object DeclarationAmendment extends DepartureRequestMessageType("IE013", "CC013C") {
+    override val officeNode: String = "CustomsOfficeOfDeparture"
```
drnaylor-oc, line 70: Put this on `DepartureRequestMessageType` directly, rather than here, you've duplicated this unnecessarily.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessageController.scala
```diff
@@ -82,7 +88,22 @@ class MessagesController @Inject() (
...
+    }
 
+  private def submit(
+    movementType: MovementType,
+    movementId: MovementId,
+    messageId: MessageId,
+    messageType: MessageType,
+    payload: Source[ByteString, _]
+  )(implicit hc: HeaderCarrier) =
+    if (isRequestMessageType(messageType)) {
```
drnaylor-oc, line 100: Rather than doing this here like this, which works against the pattern above, would it not be better to create this function here (from memory so may need tweaking, and this should be in the controller, because it's a presentation problem really):

```scala
def filterRequestMessageType(messageType: MessageType): EitherT[Future, PresentationError, RequestMessageType] = messageType match {
    case t: RequestMessageType => EitherT.rightT(t)
    case _ => EitherT.leftT(PresentationError.badRequest("messageType is not valid for requests"))
}
```

and use that in the for comprehension?

```scala
(for {
    messageType <- MessageTypeHeaderExtractor.extract(request.headers).asPresentation
    requestMessageType <- filterRequestMessageType(messageType)
    submitted   <- submit(movementType, movementId, messageId, requestMessageType, request.body)
} yield ...
```

and ditch this `submit` method.

Also note that it's a bad request that's appropriate here, because this service has done nothing wrong, it's the calling service that has.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
```diff
@@ -152,10 +164,14 @@ object MessageType {
   // ****************
 
   /** E_REQ_REL (IE054) */
-  case object ArrivalNotification extends ArrivalRequestMessageType("IE007", "CC007C")
+  case object ArrivalNotification extends ArrivalRequestMessageType("IE007", "CC007C") {
+    override val officeNode: String = "CustomsOfficeOfDestinationActual"
```
drnaylor-oc, line 168: Put this on `ArrivalRequestMessageType`

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -87,39 +88,33 @@ class RoutingServiceImpl @Inject() (messageConnectorProvider: EISConnectorProvid
...
-      case Right(office) =>
-        postToEis(office).map {
-          case Right(_)    => Right(())
-          case Left(error) => Left(error)
-        }
-      case Left(routingError) =>
-        logger.error(s"Unable to extract office of departure: $routingError")
-        Future.successful(Left(routingError));
+    val maybeACustomsOffice = {
+      import concurrent.ExecutionContext.Implicits.global
```
drnaylor-oc, line 100: Rather than importing this here, we should implicitly pass the `ExecutionContext` down.

app/uk/gov/hmrc/transitmovementsrouter/models/CustomsOffice.scala
```diff
@@ -16,7 +16,7 @@
 
 package uk.gov.hmrc.transitmovementsrouter.models
 
-case class OfficeOfDeparture(value: String) extends AnyVal {
+case class CustomsOffice(value: String) extends AnyVal {
   def isGB: Boolean = value.startsWith("GB")
-  def isXi: Boolean = !isGB
+  def isXi: Boolean = value.startsWith("XI")
```
drnaylor-oc, line 21: Nitpick: we should probably make this `isXI` for consitency on capitalisation with `isGB`.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
derekspalmer: Agree. Will move up - same for ArrivalRequestMessageType.

app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
derekspalmer: added as an implicit.

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -255,6 +257,19 @@ class MessageControllerSpec extends AnyFreeSpec with Matchers with TestActorSyst
         "message" -> "Internal server error"
       )
     }
+
+    "must return INTERNAL_SERVER_ERROR when a message is not a request message" in {
```
drnaylor-oc, line 261: BAD_REQUEST

drnaylor-oc (commented):
One minor thing, then should be good to go.

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
derekspalmer: Yeah. Good spot.

drnaylor-oc (approved):

PR #16, opened 2022-10-14
CTCP-1291: revert postToEIS

derekspalmer:


drnaylor-oc:
I don't think this is entirely the correct solution, because that restriction of ONLY GB or XI is correct.

Ultimately, if we get an office of destination that is wrong (so not GB or XI) then we should be recognising that we need to tell the user they're doing the wrong thing. I didn't think about this at the time but the error there shouldn't be unexpected, it should be returning a bad request, and this should bubble to the user.

That said, this shouldn't get this far anyway -- we might need to update the XSDs to catch this -- even then, this should be a 400 because it's the submitted data that's the problem, not this service.

drnaylor-oc:
That should have been part of the tests for this ticket - otherwise we'd always be paralysed by such changes. It's unfortunate, but we shouldn't delay correct behaviour like this.

The tests need to be fixed, not delayed.

drnaylor-oc:
https://github.com/hmrc/transit-movements-router/pull/17 also touches on this. 

I'll look at fixing the regression tests this afternoon.

drnaylor-oc:
https://github.com/hmrc/ctc-traders-api-tests/pull/41 fixed the error, so closing this.

PR #19, opened 2022-10-18
Call send message endpoint on transit-movements-push-notifications 

liandra-giacomine:


app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -0,0 +1,91 @@
...
+  ): EitherT[Future, PushNotificationError, Unit]
+}
+
+@Singleton
+class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig: AppConfig) extends PushNotificationsConnector {
+
+  val baseUrl           = appConfig.transitMovementsPushNotificationsUrl
+  val baseRoute: String = "/transit-movements-push-notifications"
+
+  private def pushNotificationMessageUpdate(movementId: MovementId, messageId: MessageId): UrlPath =
```
sonisn1988, line 62: can we move this to BaseConnector just like we are doing it for other micro services connector? BaseConnector is not available though for this .

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -0,0 +1,91 @@
...
+    ec: ExecutionContext
+  ): EitherT[Future, PushNotificationError, Unit] =
+    EitherT {
+      val url = baseUrl.withPath(pushNotificationMessageUpdate(movementId, messageId))
+
+      httpClientV2
+        .post(url"$url")
+        .addHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML)
+        .withBody(source)
+        .execute[Either[UpstreamErrorResponse, HttpResponse]]
```
sonisn1988, line 76: I think we can move this part to BaseConnector ? Just like below
    def executeAndExpect(expected: Int)(implicit ec: ExecutionContext): Future[Unit] =
      requestBuilder
        .execute[HttpResponse]
        .flatMap {
          response =>
            response.status match {
              case `expected` => Future.successful(())
              case _          => response.error
            }
        }

liandra-giacomine: This function returns an Either, if I was to use executeAndExpect which returns Future[Unit] I would need to still match on the error so I'm not sure this would make any improvement

app/uk/gov/hmrc/transitmovementsrouter/controllers/errors/ConvertError.scala
```diff
@@ -59,6 +60,15 @@ trait ConvertError {
     }
   }
 
+  implicit val pushNotificationsErrorConverter = new Converter[PushNotificationError] {
+    import uk.gov.hmrc.transitmovementsrouter.models.errors.PushNotificationError._
```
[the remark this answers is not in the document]
liandra-giacomine: They are not the same thing. The trait import at the top allows us to reference the type in the code e.g. Converter[PushNotificationError] whereas in this import we are importing all case classes present within the PushNotificationError object

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
liandra-giacomine: I am not sure whether that would bring any benefit in the router. I'm sure the BaseConnector was created for a good reason elsewhere but in this case, these values are not going to be referenced elsewhere and as of yet we don't have functionalities in common being used across the connectors.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
liandra-giacomine: and since we are not making use of a service here and just calling the connector from the controller, we need to return an eitherT

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -61,7 +60,7 @@ trait PersistenceConnector {
 }
 
 @Singleton
-class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig: AppConfig) extends PersistenceConnector with Logging {
+class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig: AppConfig) extends PersistenceConnector {
```
RobertLaverick, line 63: No change to suggest but @drnaylor-oc do you think we should consider our rules for logging regarding connectors? Feels weird to log nothing.

drnaylor-oc: Yeah, we should really. Maybe have something like we do in the EIS router for P4 in all our connectors, so we can pin-point problems in the connectors quickly, perhaps?

RobertLaverick (dismissed):

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
RobertLaverick: Lets have that discussion outside the PR!

it/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationConnectorSpec.scala
```diff
@@ -130,5 +135,16 @@ class PushNotificationConnectorSpec
       }
     }
 
+    "when the service is disabled, return a unit" in {
+      when(mockAppConfig.pushNotificationsEnabled).thenReturn(false)
+
+      stub(OK)
```
drnaylor-oc, line 141: In thie scenario, the stub should never be hit -- can we change this to `INTERNAL_SERVICE_ERROR` so that if it gets hit, we'll get a Left and know the test failed?

app/uk/gov/hmrc/transitmovementsrouter/config/CTCServicesConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+import play.api.Configuration
+import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
+
+/*
+ * This class exists simply to expose the config function, which allows us to
+ * avoid a magic string already present in the ServicesConfig object.
+ */
+class CTCServicesConfig @Inject() (config: Configuration) extends ServicesConfig(config) {
+
+  override def config(serviceName: String): Configuration = super.config(serviceName)
```
RobertLaverick, line 29: I saw this in another PR, why is this being done?

drnaylor-oc: See the comment in my PR -- there is a comment explaining why on the class.

liandra-giacomine: Explained here: https://github.com/hmrc/common-transit-convention-traders/pull/314/files/47fc965442421acf5ecde83b660bc0bd4adec5af#

drnaylor-oc (approved):

PR #24, opened 2022-11-29
[CTCP-1065] updated the endpoint url and the headers for eis

prashanthmamidi:


app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -89,10 +92,16 @@ class EISConnectorImpl(
       (t: Either[RoutingError, Unit]) => Future.successful(t.isRight),
       onFailure
     ) {
+      val data          = messageSender.value.split("-")
+      var correlationId = UUID.randomUUID().toString
```
liandra-giacomine, line 96: What's the decision around using a random uuid vs setting it as expected with movement and message id?

liandra-giacomine: Also, a more functional way to write this would be:

val correlationId = messageSender.value.split("-") match {
case 2 => CorrelationId(MovementId(data(0)), MessageId(data(1))).value
case _ => UUID.randomUUID().toString
}

using var is disencouraged

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -89,10 +92,16 @@ class EISConnectorImpl(
...
+      if (data.size == 2) {
+        correlationId = CorrelationId(MovementId(data(0)), MessageId(data(1))).value
+      }
       val requestHeaders = hc.headers(OutgoingHeaders.headers) ++ Seq(
-        "X-Correlation-Id"        -> UUID.randomUUID().toString,
+        "X-Correlation-Id"        -> correlationId,
         "CustomProcessHost"       -> "Digital",
         "X-Message-Sender"        -> messageSender.value,
```
liandra-giacomine, line 103: "X-Message-Sender" is repeated, we set it here and also is present in OutgoingHeaders.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
prashanthmamidi: updated the code - for the question, used random uuid only if the Message Sender doesn't have the data for some reason.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
prashanthmamidi: I am deleting. from OutgoingHeaders as we construct messageSender in the router app

test/uk/gov/hmrc/transitmovementsrouter/models/CorrelationIdSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+package uk.gov.hmrc.transitmovementsrouter.models
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class CorrelationIdSpec extends AnyFreeSpec with Matchers {
+
+  "build CorrelationId from MovementId and MessageId" in {
+    val correlationId = CorrelationId(MovementId("63629b56d8d9c033"), MessageId("63629b4d1d74477e"))
+    correlationId.value mustBe "63629b56-d8d9-c033-63629b4d-1d74477e"
```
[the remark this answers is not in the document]
prashanthmamidi: good spot, will update this

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
liandra-giacomine: The message sender object is constructed from the movement and message id, so it will always have the data.Also we wouldn't want to be sending a random value for that as we will be using it to match up movements.

object MessageSender {

  def apply(movementId: MovementId, messageId: MessageId): MessageSender =
    MessageSender(s"${movementId.value}-${messageId.value}")
}

Can you please just update it to:

val correlationId = CorrelationId(messageSender.movementId, messageSender.messageId)

it/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnectorSpec.scala
```diff
@@ -215,6 +217,37 @@ class EISConnectorSpec
       }
     }
```
liandra-giacomine, line 219: please check we're sending all mandatory headers

drnaylor-oc:
Replaced by https://github.com/hmrc/transit-movements-router/pull/34

PR #25, opened 2022-12-02
[CTCP-1065] updated correlation id regex pattern

prashanthmamidi:


liandra-giacomine (approved):

liandra-giacomine (commented):
before merging we need to run regression tests to make sure this change won't break anything

PR #32, opened 2023-01-23
Ctcp 2012

luceha:
Removed code which added messageSender


drnaylor-oc:
The header `X-Message-Sender` is still present in `OutgoingHeaders`, that'll need removing.

drnaylor-oc:
Compilation failed -- please take a look at what has failed and rectify that!

test/uk/gov/hmrc/transitmovementsrouter/services/XmlParserSpec.scala
```diff
@@ -153,7 +123,7 @@ class XmlParserSpec extends AnyFreeSpec with TestActorSystem with Matchers with
 
     def messageWithMessageSender(messageType: RequestMessageType): NodeSeq = {
```
liandra-giacomine, line 124: If this function isn't being used anywhere anymore, it can be removed

app/uk/gov/hmrc/transitmovementsrouter/services/XmlParser.scala
```diff
@@ -40,14 +33,6 @@ object XmlParser extends XmlParsingServiceHelpers {
...
 
-  def messageSenderWriter(messageType: MessageType, messageSender: MessageSender): Flow[ParseEvent, ParseEvent, NotUsed] = Flow[ParseEvent]
-    .mapConcat(
-      element =>
-        if (isElement(messageType.rootNode, element))
-          Seq(element, StartElement("messageSender"), Characters(s"${messageSender.value}"), EndElement("messageSender"))
-        else Seq(element)
-    )
-
   private def isElement(name: String, event: ParseEvent): Boolean = event match {
```
drnaylor-oc, line 36: One more thing, as you removed the code that used this, this is no-longer used, so this should be removed as well (it's good to make sure we don't keep things around we're not using, if we need it again, we can always get it out of the version control history).

luceha: thank you, I understand I will delete it now

drnaylor-oc (approved):

PR #33, opened 2023-02-10
CTCP-2358: Add authentication and message unwrapping

drnaylor-oc:
Note that I have removed the message size action. This is intentional, while we only expect 5MB to come back up through the small message route, we shouldn't reject anything that comes through from EIS.

I have also dropped the hybrid memory/file streaming to go for a pure file based streaming, due to the problem that Akka has some weird race conditions that they use Thread.sleep to work around in tests, so our code is actually too fast for Akka. We value reliability over speed, so we have done this.

We also have to consider the wrapped XML that EIS send. While I have written to our spec, I'm concerned that based on the us to EIS spec that we have to wrap the XML differently and that EIS will use the same wrapping. If so, the wrapping logic is now abstracted out so we just need to update the transformer -- the rest of our code works on the unwrapped XML.

Edit: I got confirmation that EIS are indeed going to send something a bit different. This is now accounted for.


app/uk/gov/hmrc/transitmovementsrouter/controllers/actions/AuthenticateEISToken.scala
```diff
@@ -0,0 +1,68 @@
...
+      } yield token) match {
+        case Some(_) => Future.successful(None)
+        case None    => Future.successful(Some(createUnauthorisedResponse(request.headers)))
+      }
+
+    } else Future.successful(None)
+
+  private def createUnauthorisedResponse(headers: Headers): Result =
+    Status(UNAUTHORIZED)(Json.toJson(PresentationError.unauthorisedError("Supplied Bearer token is invalid")))
+      .withHeaders(headers.headers.filter(_._1.toLowerCase == "x-correlation-id"): _*)
```
liandra-giacomine, line 65: could use headers.getAll("x-correlation-id")

drnaylor-oc: That returns a `Seq[String]`, not a `Seq[(String, String)]`, I need the latter!

drnaylor-oc:
This will need alterations to the test support API to work -- mostly the URL that it calls.

drnaylor-oc:
...so I hadn't realised that the conversation ID had already been sorted out because it was termed the correlation ID. This PR changes how it's retrieved (it's actually simpler now), removing the binder because I hadn't realised that it had been updated already.

So, this is ready for review.

app/uk/gov/hmrc/transitmovementsrouter/controllers/stream/StreamingParsers.scala
```diff
@@ -52,16 +56,35 @@ trait StreamingParsers extends StreamWithFile with FutureConversions {
...
-          }
+          val tempFile = temporaryFileCreator.create()
+          request.body
+            .via(transformer)
+            .runWith(FileIO.toPath(tempFile))
+            .transformWith {
+              case Success(_) => block(request.withBody(FileIO.fromPath(tempFile)))
+              case Failure(error: IOOperationIncompleteException)
+                  if error.getCause.isInstanceOf[IllegalStateException] || error.getCause.isInstanceOf[WFCException] =>
+                Future.successful(Status(BAD_REQUEST)(Json.toJson(PresentationError.badRequestError(error.getCause.getMessage))))
```
liandra-giacomine, line 74: Would it be a bad request if the transformation done by us causes the failure/how do we end up getting this failure?

drnaylor-oc: The bad request is for catching incorrect incoming XML -- those errors are caused when parsing the initial XML, so bad request is correct here.

liandra-giacomine (approved):

PR #34, opened 2023-02-14
CTCP 1896: Implement MDTP > EIS specification

drnaylor-oc:


app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -82,36 +92,48 @@ class EISConnectorImpl(
         Future.failed(new IllegalStateException(s"An unexpected error occurred - got a ${x.getClass}"))
     }
 
-  override def post(body: Source[ByteString, _], hc: HeaderCarrier): Future[Either[RoutingError, Unit]] =
+  private def extractHeaders(movementId: MovementId, messageId: MessageId, hc: HeaderCarrier): Seq[(String, String)] = {
+    val dateHeader: Seq[(String, String)] = hc.headers(Seq(HeaderNames.DATE)) match {
+      case Seq() => Seq(HeaderNames.DATE -> HTTP_DATE_FORMATTER.format(OffsetDateTime.now()))
```
liandra-giacomine, line 97: Same outcome, just wonder if this is more readable/concise? 

`      val dateHeader = Seq((HeaderNames.DATE, request.headers.get(HeaderNames.DATE).getOrElse(HTTP_DATE_FORMATTER.format(OffsetDateTime.now()))))
`

liandra-giacomine: In fact, you could just add it to the headers:
`val date = request.headers.get(HeaderNames.DATE).getOrElse(HTTP_DATE_FORMATTER.format(OffsetDateTime.now()))))`

`HeaderNames.DATE -> date`

drnaylor-oc: The issue is that we end up with duplicated headers if we just add it -- something I was trying to avoid (it was happening before this refactor). 

`getOrElse` won't work -- it returns an empty list, not an optional.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -82,36 +92,48 @@ class EISConnectorImpl(
...
+      case x     => x
+    }
+
+    val requestHeaders: Seq[(String, String)] = hc.headers(Seq("X-Message-Type")) ++ dateHeader ++ Seq(
+      "X-Correlation-Id"        -> UUID.randomUUID().toString,
```
sonisn1988, line 102: I think this  "X-Correlation-Id" and others below String better to define inside Constants.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -82,36 +92,48 @@ class EISConnectorImpl(
...
     ) {
-      val requestHeaders = hc.headers(OutgoingHeaders.headers) ++ Seq(
-        "X-Correlation-Id"        -> UUID.randomUUID().toString,
-        "CustomProcessHost"       -> "Digital",
-        HeaderNames.ACCEPT        -> MimeTypes.XML, // TODO: is this still relevant? Can't use ContentTypes.XML because EIS will not accept "application/xml; charset=utf-8"
-        HeaderNames.AUTHORIZATION -> s"Bearer ${eisInstanceConfig.headers.bearerToken}"
-      )
-
       implicit val headerCarrier: HeaderCarrier = hc
         .copy(authorization = None, otherHeaders = Seq.empty)
```
sonisn1988, line 121: is this authorization = None is required? I can see we have set that with extractHeaders. So should be removed?

app/uk/gov/hmrc/transitmovementsrouter/services/EISMessageTransformers.scala
```diff
@@ -31,27 +31,52 @@ import uk.gov.hmrc.transitmovementsrouter.services.state.UnwrappingState
...
 
 @Singleton
 class EISMessageTransformersImpl extends EISMessageTransformers {
 
+  private val WRAPPED_MESSAGE_TYPE_PREFIX = "txd"
+  private val WRAPPED_MESSAGE_ROOT_PREFIX = "n1"
+  private val UNWRAPPED_PREFIX            = "ncts"
+  private val NCTS_NAMESPACE_URL          = "http://ncts.dgtaxud.ec"
+  private val XSI_NAMESPACE_URL           = "http://www.w3.org/2001/XMLSchema-instance"
+  private val EIS_NAMESPACE_URL           = "http://www.hmrc.gov.uk/eis/ncts5/v1"
```
sonisn1988, line 53: I Think we can move all this to Constants to make this class more clean and clear.

drnaylor-oc: I disagree. These constants are only relevant here so I think they should remain here.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
drnaylor-oc: Yes, I was seeing duplication of headers when I left it in.

it/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnectorSpec.scala
```diff
@@ -129,43 +146,104 @@ class EISConnectorSpec
...
             )
               .inScenario("Standard Call")
               .whenScenarioStateIs(currentState)
+              .withHeader("Date", anything)
               .withHeader("X-Correlation-Id", matching(RegexPatterns.UUID))
+              .withHeader("X-Conversation-Id", equalTo(expectedConversationId.value.toString))
               .withHeader("CustomProcessHost", equalTo("Digital"))
               .withHeader(HeaderNames.ACCEPT, equalTo("application/xml"))
+              .withHeader(HeaderNames.CONTENT_TYPE, equalTo("application/xml"))
               .willReturn(aResponse().withStatus(codeToReturn))
```
sonisn1988, line 164: Not sure. But do we need to test with some dummy Authorisation header?

drnaylor-oc: This is a bit of a weird one to think about, but the answer is no.

What we're doing in the tests in this service is testing how our system reacts to certain scenarios. In these integration tests, we are saying "we expect these headers, with these values". What we're really doing that is testing that the receiving function is getting what we asked for -- in this case, we have tested that the token we have set in the config has ended up hitting the dummy endpoint.

Now, we define how the endpoint behaves, so in the lines you highlighted, we say "if we get these headers, then respond in this way". We can respond how we like to test the behaviour to the response. 

So, we don't need to test with a dummy header, because that adds nothing -- we know that whatever we put into the config is output here. We need to test how different responses are handled, and do it in the simplest way -- in this case, same input, different response.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
drnaylor-oc: I'm somewhat open to this because this could benefit the tests too -- I may put these in a companion object rather than a constants object elsewhere though but they have no value outside of this connector.

drnaylor-oc: I've created a `RouterHeaderNames` class, but I've not used it in the tests -- as the tests that enforce a spec should be separate from the production code as much as possible, including constants.

drnaylor-oc:
Thankfully, this works with the stub as is, so will merge this. Will tighten up the stub in the other PR.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
liandra-giacomine: Ah! Gotcha, thanks :)

PR #35, opened 2023-02-14
Ctcp 2022: Create callback endpoint for receiving a message to/from Upscan

liandra-giacomine:


drnaylor-oc:
There are conflicts... while you might want to fix them please hold off from merging it if it gets approved for the moment, thanks!

drnaylor-oc:
We can continue with this, we'll likely need upscan after all for a few reasons given by MDTP.

sonisn1988 (dismissed):
There are few conflict. Please resolve that.

test/uk/gov/hmrc/transitmovementsrouter/services/UpscanServiceSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import play.api.libs.json.Json
+import uk.gov.hmrc.transitmovementsrouter.generators.TestModelGenerators
+
+class UpscanServiceSpec extends AnyFreeSpec with ScalaFutures with Matchers with ScalaCheckPropertyChecks with TestModelGenerators {
+
+  val UpscanService = new UpscanServiceImpl
```
sonisn1988, line 28: small correction I suggest use upscanService

app/uk/gov/hmrc/transitmovementsrouter/models/responses/UpscanResponse.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object FailureDetails {
+  implicit val format = Json.format[FailureDetails]
+}
+
+object UpscanResponse {
+
+  final case class SuccessfulSubmission(reference: String, downloadUrl: String, fileStatus: String, uploadDetails: UploadDetails) extends UpscanResponse
+
+  final case class SubmissionFailure(reference: String, fileStatus: String, failureDetails: FailureDetails) extends UpscanResponse
```
sonisn1988, line 41: I think we can maintain fileStatus like below?
sealed trait UpscanFileStatus

object UpscanFileStatus {
  case object Ready extends UpscanFileStatus
  case object Failed extends UpscanFileStatus

sonisn1988, line 41: reference could be final case class UpscanReference(value: String) extends AnyVal

liandra-giacomine (commented):
Move upscan deserialisation to controlle

app/uk/gov/hmrc/transitmovementsrouter/models/responses/UpscanResponse.scala
liandra-giacomine: Turned it into a case class instead, just worried about hard coding it

drnaylor-oc: I don't think we should worry so much about it -- I don't expect this to change. Because the schema changes based on this value, it's probably not going to change. If it does, we might have larger issues with how the schema looks anyway.

app/uk/gov/hmrc/transitmovementsrouter/controllers/UpscanResponseParser.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import play.api.Logging
+import play.api.libs.json.JsValue
+import play.api.mvc.BaseController
+import uk.gov.hmrc.transitmovementsrouter.models.responses.UpscanResponse
+
+trait UpscanResponseParser {
+  self: BaseController with Logging =>
+
+  def parseAndLogUpscanResponse(responseBody: JsValue): Option[UpscanResponse] =
```
sonisn1988, line 27: I think better to move this method inside controller and change method name to validateUpscanResponse.

app/uk/gov/hmrc/transitmovementsrouter/models/errors/UpscanCallbackError.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovementsrouter.models.errors
+
+sealed trait UpscanCallbackError
+
+object UpscanCallbackError {
+  case object UnexpectedResponse extends UpscanCallbackError
```
sonisn1988, line 22: I can't see any usage of this Error also it is not mapped with ConvertError. if not using then please remove it.

app/uk/gov/hmrc/transitmovementsrouter/controllers/UpscanResponseParser.scala
drnaylor-oc: I'm non-committal about it. I can see why we'd have it as a trait, if only for testing it'll potentially make things easier. I'd probably put it in the controller myself, but it's not a dealbeaker for me. 

Do we need the `BaseController` self type?

Also, rather than returning an option, we know that this will return an appropriate error message at some point. I think this should be returning `EitherT[Future, PresentationError, UpscanResponse]` and return a bad request if we  got a bad response from Upscan.

Leave actually processing the response for success/failure for now. I envisage that this function will be split into two, validate (create the business object) and process (use the business object), the latter probably going into a service.

liandra-giacomine: @drnaylor-oc  The request will coming from upscan, I thought that our response back to them would be irrelevant, which is why I didn't implement an error in an eitherT (I did at first, then realised there's no point?) but perhaps I'm misunderstanding something

app/uk/gov/hmrc/transitmovementsrouter/models/errors/UpscanCallbackError.scala
liandra-giacomine: @drnaylor-oc  Created to be wrapped in an eitherT then realised we wouldn't be returning an error status code to upscan, I'll keep it for now in case my understanding is wrong

app/uk/gov/hmrc/transitmovementsrouter/controllers/UpscanResponseParser.scala
drnaylor-oc: >  The request will coming from upscan, I thought that our response back to them would be irrelevant, which is why I didn't implement an error in an eitherT (I did at first, then realised there's no point?) but perhaps I'm misunderstanding something

If they've sent us something that doesn't fit their schema/contract (broken Json, or a field that no longer exists, for example), we should tell them by sending a 400. If we get a "file failed" callback with a valid failure schema then _that's_ still a 2xx, because we'll handle the failure successfully.

> I made the BaseController a self type because of the issue you pointed out regarding this functionality belonging in the controller/presentation layer.

Okay! I just thought it might be a bit of a pain from a testing view, but that makes sense. Not too bothered about removing it.

app/uk/gov/hmrc/transitmovementsrouter/models/errors/UpscanCallbackError.scala
drnaylor-oc: As above. Note that because the trait is concerned with the presentation layer only, then we should just use a `PresentationError` directly, so this is still superfluous.

app/uk/gov/hmrc/transitmovementsrouter/models/responses/UpscanResponse.scala
liandra-giacomine: Ok

app/uk/gov/hmrc/transitmovementsrouter/models/responses/UpscanResponse.scala
```diff
@@ -0,0 +1,104 @@
...
+  object Reference {
+
+    implicit val format: Format[Reference] =
+      Format(
+        Reads.of[String].map(Reference(_)),
+        Writes(
+          ref => JsString(ref.value)
+        )
+      )
+  }
```
drnaylor-oc, line 57: Doesn't necessarily need changing now, but this value type conversion could be replaced with the macro:

```suggestion
    implicit val format: Format[Reference] = Json.valueFormat[Reference]
```

drnaylor-oc (dismissed):
I don't have any issues with this going in, but something for note in the future to speed up development a little.

sonisn1988 (approved):
LGTM

PR #37, opened 2023-02-28
Send large message metadata to transit-movements

sonisn1988:


app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -69,17 +72,32 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+      ((source, objectStoreURI) match {
+        case (Some(src), _) =>
+          httpClient.transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML, RouterHeaderNames.MESSAGE_TYPE -> messageType.code)).withBody(src)
+        case (_, Some(objectStoreURIHeader)) =>
+          httpClient.transform(
+            _.addHttpHeaders(
+              RouterHeaderNames.MESSAGE_TYPE     -> messageType.code,
+              RouterHeaderNames.OBJECT_STORE_URI -> objectStoreURIHeader.value
+            )
+          )
```
drnaylor-oc, line 99: So what happens when you have neither or both? Your function suggests that's possible when it shouldn't be. Just because we don't call it that way, doesn't mean that someone in live services who takes over in a years' time will understand that.

This is too complicated for what it needs to be. While I'm a big fan of "Don't Repeat Yourself", I'm also a fan of keeping things simple. Really, you're calling this endpoint in two very distinct ways.

Why not make two functions like so:

```
  def postBody(
    movementId: MovementId,
    messageId: MessageId,
    messageType: MessageType,
    source: Source[ByteString, _]
  )(implicit
    hc: HeaderCarrier,
    ec: ExecutionContext
  ): EitherT[Future, PersistenceError, PersistenceResponse]

  def postObjectStoreUri(
    movementId: MovementId,
    messageId: MessageId,
    messageType: MessageType,
    objectStoreURI: ObjectStoreURI
  )(implicit
    hc: HeaderCarrier,
    ec: ExecutionContext
  ): EitherT[Future, PersistenceError, PersistenceResponse]
```

an in each case set up the client. Creating the URL and the http client could be done in its own function too if you wanted. Once that's done, you can have a third method that takes the client and then does the common execution part. That way it's clear:

* These are two different calls for two different cases
* You must have one of the body, or the object store URL
* The response is the same regardless

No point shoehorning all this into one function if it makes this less clear.

drnaylor-oc (commented):
Let's not make things more complicated then they need to be.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
sonisn1988: Done the changes as suggested above. Please re-review.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -105,8 +103,9 @@ class PersistenceConnectorSpec
       val body = Json.obj("messageId" -> messageId.value).toString()
 
       stub(OK, Some(body))
-
-      whenReady(connector.post(movementId, messageId, DeclarationAmendment, source).value) {
+      implicit val hc: HeaderCarrier =
+        HeaderCarrier(extraHeaders = Seq("X-Message-Type" -> MessageType.DeclarationAmendment.code, HeaderNames.CONTENT_TYPE -> MimeTypes.XML))
```
drnaylor-oc, line 107: You're partially defeating the object of the test here. Why are you adding these here when we should be testing that the connector is adding them?

We should probably also update the wiremock stub to check for appropriate headers and bodies -- after all, that's the whole point of this ticket.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -118,8 +117,9 @@ class PersistenceConnectorSpec
         server.resetAll()
 
         stub(statusCode)
-
-        whenReady(connector.post(movementId, messageId, DeclarationAmendment, source).value) {
+        implicit val hc: HeaderCarrier =
+          HeaderCarrier(extraHeaders = Seq("X-Message-Type" -> MessageType.DeclarationAmendment.code, HeaderNames.CONTENT_TYPE -> MimeTypes.XML))
```
drnaylor-oc, line 121: Same here

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -139,8 +139,62 @@ class PersistenceConnectorSpec
...
+  }
+
+  "post for Large Message" should {
+    "return messageId when post is successful" in {
+      val body           = Json.obj("messageId" -> messageId.value).toString()
+      val objectStoreURI = s"common-transit-convention-traders/movements/$movementId/x-conversion-id.xml"
+      stub(OK, Some(body))
+      implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders =
+        Seq("X-Message-Type" -> MessageType.DeclarationAmendment.code, "X-Object-Store-Uri" -> ObjectStoreURI(objectStoreURI).value)
+      )
```
drnaylor-oc, line 160: And here too. In fact, this is worse because I would NOT expect to have the `X-Object-Store-Uri` or `X-Message-Type` header in the header carrier at all (in the previous case, we'd only have the `Content-Type`). Additionally, because you're relying on the lack of a content type at the target, then we really need to be checking for that too.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -139,8 +139,62 @@ class PersistenceConnectorSpec
...
+    }
+
+    "return a PersistenceError when unsuccessful" in forAll(errorCodes) {
+      statusCode =>
+        server.resetAll()
+        val objectStoreURI = s"common-transit-convention-traders/movements/$movementId/x-conversion-id.xml"
+        stub(statusCode)
+        implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders =
+          Seq("X-Message-Type" -> MessageType.DeclarationAmendment.code, "X-Object-Store-Uri" -> ObjectStoreURI(objectStoreURI).value)
+        )
```
drnaylor-oc, line 175: And here

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -139,8 +139,62 @@ class PersistenceConnectorSpec
...
+    }
 
-      whenReady(connector.post(movementId, messageId, DeclarationAmendment, failingSource).value) {
+    "return Unexpected(throwable) when NonFatal exception is thrown" in {
+      server.resetAll()
+      val objectStoreURI = s"common-transit-convention-traders/movements/$movementId/x-conversion-id.xml"
+      stub(OK)
+      implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders =
+        Seq("X-Message-Type" -> MessageType.DeclarationAmendment.code, "X-Object-Store-Uri" -> ObjectStoreURI(objectStoreURI).value)
+      )
```
drnaylor-oc, line 196: And here.

drnaylor-oc (commented):
These tests aren't testing anything new, and actually might miss something because of how you've set it up.

This code introduces a second path for submitting a message to `transit-movements`, by the body as a payload, and the object-store URI as a header. None of your tests check that the Wiremock server has been hit with the correct headers (or absence of the `Content-Type` header), and in fact you pre-seed the `HeaderCarrier` with the headers the connector should be adding, so they're always going to be there and mask any issues we might have.

Please take a look at the tests again and make sure the Wiremock stub in each case is verifying that we have the right headers in place, and that in the case of the large message route that we have no body.

(To add: the original tests looks like they were written by me, and I should have been testing for the headers in the first place -- so that's a mistake there by me... but we really need to be a lot better about these things, particularly in this scenario).

drnaylor-oc:
You need to ensure you call this in the `incomingLargeMessage` method in the controller too, and set up all the error handling and such there.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
sonisn1988: I agree and Done the changes as suggested

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -69,34 +82,56 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+            case NOT_FOUND             => Left(MovementNotFound(movementId))
+            case INTERNAL_SERVER_ERROR => Left(Unexpected())
+            case _                     => Left(Unexpected())
+          }
+      }
+      .recover {
+        case NonFatal(ex) => Left(Unexpected(Some(ex)))
+      }
+
+  private def requestUrl(movementId: MovementId, messageId: MessageId)(implicit hc: HeaderCarrier, ec: ExecutionContext) = {
```
drnaylor-oc, line 132: nitpick: this is a bad name - it's not returning the request URL, it's returning a request builder. This should really be named "createRequest".

drnaylor-oc (commented):
One nitpick, but naming is something we should try to be good on.

drnaylor-oc (approved):

PR #41, opened 2023-03-10
changes for CTCP-2594 and CTCP-2431

sonisn1988:


app/uk/gov/hmrc/transitmovementsrouter/models/ObjectStoreRetentionPeriod.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovementsrouter.models
+
+import uk.gov.hmrc.objectstore.client.RetentionPeriod
+
+case class ObjectStoreRetentionPeriod(value: RetentionPeriod) extends AnyVal
```
drnaylor-oc, line 21: Why are you wrapping this in an `AnyVal`? `RetentionPeriod` is already a specific type so there is no need to wrap it. We wrap `String`s in `AnyVal`s because they are easy to put in the wrong place.

Remove this `AnyVal` and just keep the `RetentionPeriod`.

app/uk/gov/hmrc/transitmovementsrouter/models/ObjectStoreFileDirectory.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovementsrouter.models
+
+import uk.gov.hmrc.objectstore.client.Path
+
+case class ObjectStoreFileDirectory(value: Path.File) extends AnyVal
```
drnaylor-oc, line 21: Similarly, `Path.File` is already a specific type for object-store. No need to do this.

Remove this `AnyVal`

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -129,4 +181,9 @@ class MessagesController @Inject() (
     case _                     => EitherT.leftT(PresentationError.badRequestError(s"${messageType.code} is not valid for requests"))
   }
 
+  private def fileFormat(clock: Clock) = {
+    val dateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss").withZone(ZoneOffset.UTC)
```
drnaylor-oc, line 185: Rather than re-create this every time, put this field on the class level.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -101,9 +142,20 @@ class MessagesController @Inject() (
...
+          .addMessage(
+            downloadUrl,
+            movementId,
+            messageId,
+            ObjectStoreFileDirectory(
+              Path.File(s"movements/${movementId.value}/${movementId.value}-${messageId.value}-${fileFormat(clock)}.xml")
+            ),
+            ObjectStoreOwner("common-transit-convention-traders"),
+            ObjectStoreRetentionPeriod(RetentionPeriod.SevenYears)
+          )
```
drnaylor-oc, line 157: I don't like what you've done here -- you've leaked object-store specific parameters into the controller, when really we should be pushing this sort of thing into the service. I don't think the controller should care about the details of where object-store is putting stuff.

I suspect you did this because you were trying to make a more generic "addMessage" call, but really, all you've done is move all the decisions here, where they shouldn't be. I would much rather there be two functions on the service that set the parameters for us, something like `storeOutgoing` and `storeIncoming`, or something a bit more than that.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -64,9 +74,40 @@ class MessagesController @Inject() (
...
+              messageId,
+              ObjectStoreFileDirectory(Path.File(s"sdes/" + objectStoreResourceLocation.value)),
+              ObjectStoreOwner("transit-movements-router"),
+              ObjectStoreRetentionPeriod(RetentionPeriod.OneWeek)
+            )
+            .asPresentation
+        } yield objectSummary).fold[Result](
+          error => Status(error.code.statusCode)(Json.toJson(error)),
+          _ => Accepted
+        )
```
drnaylor-oc, line 107: As I said for the incoming, this is leaking too much here.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -64,9 +74,40 @@ class MessagesController @Inject() (
...
+    Action.async {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+        (for {
+          messageType                 <- messageTypeExtractor.extractFromHeaders(request.headers).asPresentation
+          requestMessageType          <- filterRequestMessageType(messageType)
+          objectStoreResourceLocation <- extractObjectStoreURIHeader(request.headers)
+          objectSummary <- objectStoreService
+            .addMessage(
+              DownloadUrl(s"${appConfig.objectStoreUrl}/${request.headers.get(RouterHeaderNames.OBJECT_STORE_URI).get}"),
```
drnaylor-oc, line 96: `request.headers.get(RouterHeaderNames.OBJECT_STORE_URI).get` -- isn't this just `objectStoreResourceLocation`?

sonisn1988: request.headers.get(RouterHeaderNames.OBJECT_STORE_URI).get this will include ``` common-transit-convention-traders ``` and that is required as part of download url. but objectStoreResourceLocation does not contain ``` common-transit-convention-traders ``` as we are stripping `` common-transit-convention-traders ``` in extractor.

drnaylor-oc: So, should `objectStoreResourceLocation` contain two fields, one with the context path and one without, rather than having to pull down the header twice?

sonisn1988: yeah. we can do that way. No problem.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
sonisn1988: if we defining storeOutgoing and storeIncoming then we need to make 2 calls to ObjectStoreServiceservice one for to set parameters in storeOutgoing or storeIncoming and then call addMessage. I hope you are fine with that?

drnaylor-oc: Why can't whatever `addMessage` does be called by the `store*` methods?

app/uk/gov/hmrc/transitmovementsrouter/models/ObjectStoreRetentionPeriod.scala
```diff
@@ -18,4 +18,4 @@ package uk.gov.hmrc.transitmovementsrouter.models
 
 import uk.gov.hmrc.objectstore.client.RetentionPeriod
 
-case class ObjectStoreRetentionPeriod(value: RetentionPeriod) extends AnyVal
+case class ObjectStoreRetentionPeriod(value: RetentionPeriod)
```
drnaylor-oc, line 21: I meant just get rid of this case class -- `RetentionPeriod` is specific enough as it is that it doesn't need wrapping.

app/uk/gov/hmrc/transitmovementsrouter/models/ObjectStoreFileDirectory.scala
```diff
@@ -18,4 +18,4 @@ package uk.gov.hmrc.transitmovementsrouter.models
 
 import uk.gov.hmrc.objectstore.client.Path
 
-case class ObjectStoreFileDirectory(value: Path.File) extends AnyVal
+case class ObjectStoreFileDirectory(value: Path.File)
```
drnaylor-oc, line 21: Same here.

olakunlekuye2 (approved):

PR #44, opened 2023-03-29
CTCP-2433: Submit file to SDES (transit-movements-router)

sonisn1988:


conf/application.conf
```diff
@@ -186,6 +193,19 @@ microservice {
...
+      host = localhost
+      port = 8764
+      file-ready-uri = "/notification/fileready"
+    }
+
+    sdes-stub {
+       host = localhost
+       port = 9191
+       file-ready-uri = "/sdes-stub/notification/fileready"
+    }
```
drnaylor-oc, line 207: This shouldn't exist. Stub details should replace those of the proxy.

I imagine what you've done is copy the message parser. You don't need to do that here, because I did that to ensure that we could use the proxy properly. I wouldn't worry about that here.

app/uk/gov/hmrc/transitmovementsrouter/config/AppConfig.scala
```diff
@@ -46,4 +48,23 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
   lazy val objectStoreUrl: String =
     config.get[String]("microservice.services.object-store.sdes-host")
 
+  // SDES configuration
+  lazy val useProxy: Boolean = config.get[Boolean]("sdes.use-proxy") // set true for sdes proxy and false for sdes stub
+  lazy val service: String   = if (useProxy) "secure-data-exchange-proxy" else "sdes-stub"
```
drnaylor-oc, line 53: No need for this -- if you're just switching between the two just put the configuration in `secure-data-exchange-proxy`.

olakunlekuye2 (approved):

PR #47, opened 2023-04-11
CTCP-2434 Sdes callback message status update

akila-sankar:


app/uk/gov/hmrc/transitmovementsrouter/controllers/SdesResponseParser.scala
```diff
@@ -0,0 +1,58 @@
...
+import play.api.Logging
+import play.api.libs.json.JsValue
+import play.api.mvc.BaseController
+import uk.gov.hmrc.transitmovementsrouter.controllers.errors.PresentationError
+import uk.gov.hmrc.transitmovementsrouter.models.sdes.SdesNotificationItem
+
+trait SdesResponseParser {
+  self: BaseController with Logging =>
+
+  def parseAndLogSdesResponse(responseBody: JsValue): Either[PresentationError, SdesNotificationItem] =
```
sonisn1988, line 28: Please use EitherT[Future, PresentationError, SdesNotificationItem] as this is going to be async call. so better to use EitherT with Future

app/uk/gov/hmrc/transitmovementsrouter/controllers/SdesResponseParser.scala
```diff
@@ -0,0 +1,58 @@
...
+      .getOrElse {
+        logger.error("Unable to parse unexpected response from SDES")
+        Left(PresentationError.badRequestError("Unexpected SDES callback response"))
+      }
+
+  private def evaluate(sdesResponse: SdesNotificationItem) =
+    sdesResponse match {
+      case SdesNotificationItem(_, _, _, _, _, _, None, _, _) =>
+        logger.info(
+          s"Received a successful response from SDES callback for the following x-conversation-id: ${sdesResponse.conversationId}.get.value"
```
sonisn1988, line 46: This is not going to print  conversationId  ``` ${sdesResponse.conversationId}.get.value ```  so little bit need to modify  ${sdesResponse.conversationId.get.value} . Please do it others places as well

app/uk/gov/hmrc/transitmovementsrouter/controllers/SdesResponseParser.scala
```diff
@@ -0,0 +1,58 @@
...
+  self: BaseController with Logging =>
+
+  def parseAndLogSdesResponse(responseBody: JsValue): Either[PresentationError, SdesNotificationItem] =
+    responseBody
+      .validate[SdesNotificationItem]
+      .map {
+        sdesResponse =>
+          if (sdesResponse.conversationId.isDefined)
+            evaluate(sdesResponse)
+          else Left(PresentationError.badRequestError("Unexpected SDES callback response"))
```
sonisn1988, line 35: I think we are duplicating it here. if **conversationId**  is not defined then it will go to getOrElse condition so better to remove this else condition.

conf/app.routes
```diff
@@ -2,7 +2,8 @@
 
 POST        /traders/:eori/movements/:movementType/:movementId/messages/:messageId             uk.gov.hmrc.transitmovementsrouter.controllers.MessagesController.outgoing(eori: EoriNumber, movementType: MovementType, movementId: MovementId, messageId: MessageId)
 
-POST        /movements/:movementId/messages/:messageId                                                uk.gov.hmrc.transitmovementsrouter.controllers.MessagesController.incomingLargeMessage(movementId: MovementId, messageId: MessageId)
+POST        /movements/:movementId/messages/:messageId                                         uk.gov.hmrc.transitmovementsrouter.controllers.MessagesController.incomingLargeMessage(movementId: MovementId, messageId: MessageId)
 
 POST        /movements/:ids/messages                                                           uk.gov.hmrc.transitmovementsrouter.controllers.MessagesController.incoming(ids: ConversationId)
 
+POST        /movements/sdes/callback                                                           uk.gov.hmrc.transitmovementsrouter.controllers.MessagesController.handleSdesResponse()
```
sonisn1988, line 9: @drnaylor-oc  I think this should be PATCH call as we need to change only status with db.  What do you think?

drnaylor-oc: SDES will only send a POST.

app/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotification.scala
```diff
@@ -0,0 +1,47 @@
...
+
+sealed trait SdesNotification
+
+object SdesNotification {
+  case object FileReady             extends SdesNotification
+  case object FileReceived          extends SdesNotification
+  case object FileProcessed         extends SdesNotification
+  case object FileProcessingFailure extends SdesNotification
+
+  val values = Seq(FileProcessed, FileProcessingFailure)
```
sonisn1988, line 34: I can't see any references of this. If not needed then please remove it.

app/uk/gov/hmrc/transitmovementsrouter/controllers/SdesResponseParser.scala
drnaylor-oc: Not as written. The compiler will (currently) throw an error if the `else` block isn't defined.

We could do the following (from memory, but should show a pattern we can use):

```scala
responseBody
    .validate[SdesNotificationItem]
    .filter(_.sdesResponse.conversationId.isDefined)
    .map(evaluate)
    .getOrElse {
        ...
    }
```

app/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotification.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  implicit val writes = new Writes[SdesNotification] {
+    def writes(status: SdesNotification) = Json.toJson(status.toString())
+  }
```
sonisn1988, line 38: Please change the variable name ``` status ``` to something else like ``` sdesNotification ```

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -164,4 +170,45 @@ class MessagesController @Inject() (
...
+  )(implicit
+    hc: HeaderCarrier
+  ) = {
+    val (movementId, messageId) = extractMovementMessageId(sdesResponse)
+    (for {
+      persistenceResponse <- persistenceConnector
+        .patchMessageStatus(movementId, messageId, MessageUpdate(messageStatus))
+        .asPresentation
+
+    } yield persistenceResponse).fold[Result](
```
sonisn1988, line 185: Please yield and fold it to inside original method call as we need to implement ppns method call as well so better to do this with ``` handleSdesResponse ```

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -108,6 +124,24 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+    EitherT(
+      executeAndExpect(
+        updateStatusRequest(movementId, messageId)
+          .transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.JSON))
+          .withBody(Json.toJson(body)),
+        OK
+      )
+        .map(Right(_))
+        .recover {
+          case UpstreamErrorResponse(_, NOT_FOUND, _, _) => Left(MovementNotFound(movementId))
```
sonisn1988, line 140: Here we will be having movementId and messageId. So please create new error type inside PersistenceError like ``` case class MessageNotFound(movementId: MovementId, messageId: MessageId)        extends PersistenceError ``` then map it with ConvertError and use that here.

app/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotificationItem.scala
```diff
@@ -0,0 +1,44 @@
...
+  checksum: String,
+  availableUntil: Instant,
+  failureReason: Option[String],
+  dateTime: Instant,
+  properties: Seq[SdesProperties]
+) {
+
+  val conversationId = properties
+    .find(
+      p => p.name.toLowerCase() == RouterHeaderNames.CONVERSATION_ID.toLowerCase()
```
sonisn1988, line 38: I think here we can remove ```toLowerCase()``` from both the side  and find with actual match

drnaylor-oc: As a note, we don't know exactly what the property name is but it _will_ be case sensitive given it'll be part of a Json schema. So, for that reason alone, we should remove the `toLowerCase`.

We just don't know what that property will be yet.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -164,4 +170,45 @@ class MessagesController @Inject() (
...
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+        parseAndLogSdesResponse(request.body) match {
+          case Left(presentationError: PresentationError) => Future.successful(Status(presentationError.code.statusCode)(Json.toJson(presentationError)))
+          case Right(sdesResponse) =>
+            sdesResponse.notification match {
+              case SdesNotification.FileProcessed =>
+                updateMessageStatus(sdesResponse, MessageStatus.Success)
+              case SdesNotification.FileProcessingFailure =>
+                updateMessageStatus(sdesResponse, MessageStatus.Failed)
```
drnaylor-oc, line 209: The result of this shouldn't be ignored.

Unlike Upscan, where it only needs to tell us about a file and that's it, it doesn't care after that, SDES is actually part of the message flow so should expect a 500, which signals it to retry later. 

There are just returning `Ok`, even if our call fails we're telling SDES that everything is fine, when it's not. I do understand that for Upscan we do swallow (most) errors, but for SDES, we want it to retry if we can't update the status.

Take this comment with the one that Sandeep made above. `updateMessageStatus(...)` should be part of a for comprehension and if it fails, we tell SDES that with an internal server error.

akila-sankar: In the ticket it is mentioned as we should always be returning a 200 unless the message is malformed and I followed the same here.

If it needs to changed then fine to take it up, please confirm

drnaylor-oc: I added a comment to the ticket, but to summarise:

> If something fails (as in, internal server error, e.g. could not connect to the database), **we should return a 500 to SDES to prompt it to retry** rather than ignore it and return a 2xx error.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -112,6 +117,15 @@ class PersistenceConnectorSpec
       )
   )
 
+  def stubForPatchMessageStatus(codeToReturn: Int, body: Option[String] = None) = server.stubFor(
+    patch(
+      urlEqualTo(uriStatus)
```
sonisn1988, line 122: Use ``` .withRequestBody(equalToJson("pass the json response of sdes")) ``` with this stub to make it real stub

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -112,6 +117,15 @@ class PersistenceConnectorSpec
...
+    ).willReturn(
+      if (body.isEmpty) aResponse().withStatus(codeToReturn)
+      else aResponse().withStatus(codeToReturn)
```
sonisn1988, line 125: Pease use ``` else aResponse().withStatus(codeToReturn).withBody(body.get) ``` to test with error response with body

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -123,4 +131,49 @@ trait TestModelGenerators extends BaseGenerators {
     )
   }
 
+  implicit def arbitrarySdesResponse(): Arbitrary[SdesNotificationItem] = Arbitrary {
+    for {
+      filename          <- Gen.alphaNumStr
+      correlationId     <- Gen.alphaNumStr
+      checksum          <- Gen.stringOfN(4, Gen.alphaChar)
+      checksumAlgorithm <- Gen.alphaNumStr
+      notification = SdesNotification.FileProcessed
```
sonisn1988, line 140: use Gen.oneOf() to test it for FileProcessed and FileFailed

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -123,4 +131,49 @@ trait TestModelGenerators extends BaseGenerators {
...
+      received     = Instant.now()
+      properties   = Seq(SdesProperties("x-conversation-id", "123e4567-e89b-12d3-a456-426614174000"))
```
sonisn1988, line 142: Please use Generator here instead of hardcoding with some value

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -123,4 +131,49 @@ trait TestModelGenerators extends BaseGenerators {
...
+    )
+  }
+
+  implicit def arbitrarySdesFailureResponse(): Arbitrary[SdesNotificationItem] = Arbitrary {
+    for {
+      filename          <- Gen.alphaNumStr
+      correlationId     <- Gen.alphaNumStr
+      checksum          <- Gen.stringOfN(4, Gen.alphaChar)
+      checksumAlgorithm <- Gen.alphaNumStr
+      notification = SdesNotification.FileProcessingFailure
```
sonisn1988, line 162: in both cases (FileProcessed and FileFailed) we need to return 200 so better to use above one and remove this.

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -36,6 +39,11 @@ import java.time.format.DateTimeFormatter
 
 trait TestModelGenerators extends BaseGenerators {
 
+  lazy val genShortUUID: Gen[String] = Gen.long.map {
```
sonisn1988, line 42: if not using anywhere then remove it.

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -1086,4 +1088,58 @@ class MessageControllerSpec
         status(result) mustBe INTERNAL_SERVER_ERROR
     }
   }
+
+  "POST SDES callback" - {
+
+    lazy val movementId = MovementId("123e4567e89b12d3")
```
sonisn1988, line 1094: Use these two with forAll and use arbitrary instead of checking with hardcoded value

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
drnaylor-oc: That doesn't make sense, considering that the `body` variable is for the request, not the response.

drnaylor-oc: Yes, this `if` statement is pointless here.

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -123,4 +131,49 @@ trait TestModelGenerators extends BaseGenerators {
     )
   }
 
+  implicit def arbitrarySdesResponse(): Arbitrary[SdesNotificationItem] = Arbitrary {
```
sonisn1988, line 134: as we are not passing any parameter with this def so please make it **val** instead of **def**

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
akila-sankar: Added for this scenario

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
akila-sankar: Yes you are right, but accordingly status should be mapped dynamically in the tests and have updated it like that now

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
akila-sankar: From the upstream have scenario only for MovementNotFound and thats why only added for that alone here.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -164,4 +170,30 @@ class MessagesController @Inject() (
...
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+        (for {
+          sdesResponse <- parseAndLogSdesResponse(request.body)
+          (movementId, messageId) = extractMovementMessageId(sdesResponse)
+          messageStatus = sdesResponse.notification match {
+            case SdesNotification.FileProcessed         => MessageStatus.Success
+            case SdesNotification.FileProcessingFailure => MessageStatus.Failed
+          }
+          persistenceResponse <- persistenceConnector
```
sonisn1988, line 191: This piece of code will get executed for FileReceived and FileReady scenarios as well. Also can not see logic to ignore for FileReceived and FileReady response from SDES?

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
sonisn1988: we do have flexibility to change the error here so good to change as suggested.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
```diff
@@ -126,12 +158,29 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
           }
       }
       .recover {
-        case NonFatal(ex) => Left(Unexpected(Some(ex)))
+        case NonFatal(ex) =>
+          Left(Unexpected(Some(ex)))
+      }
+
+  def executeAndExpect(requestBuilder: RequestBuilder, expected: Int)(implicit ec: ExecutionContext) =
```
sonisn1988, line 165: make this private method

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -123,4 +127,25 @@ trait TestModelGenerators extends BaseGenerators {
...
 
+  implicit def arbitrarySdesResponse(conversationId: ConversationId): Arbitrary[SdesNotificationItem] = Arbitrary {
+    for {
+      filename          <- Gen.alphaNumStr
+      correlationId     <- Gen.alphaNumStr
+      checksum          <- Gen.stringOfN(4, Gen.alphaChar)
+      checksumAlgorithm <- Gen.alphaNumStr
+      notification      <- Gen.oneOf(SdesNotification.values)
+      received   = Instant.now()
+      properties = Seq(SdesProperties(RouterHeaderNames.CONVERSATION_ID, conversationId.value.toString))
```
sonisn1988, line 138: better to create conversationId gen here with this gen instead of creating and passing it in different test cases as parameters.

akila-sankar: I don't see a problem over here as the conversationId using should match with movementId and messageId

app/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnector.scala
akila-sankar: As discussed, it can be mapped to either one and changing it to MessageNotFound

app/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotificationItem.scala
```diff
@@ -0,0 +1,44 @@
...
+  checksum: String,
+  availableUntil: Instant,
+  failureReason: Option[String],
+  dateTime: Instant,
+  properties: Seq[SdesProperties]
+) {
+
+  val conversationId = properties
+    .find(
+      p => p.name == RouterHeaderNames.CONVERSATION_ID
```
sonisn1988, line 38: have you tested this with end to end? I think post changes merged by Dan today morning. It is not going to work

akila-sankar: Yes updated according to the recent changes

sonisn1988 (approved):
LGTM

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
```diff
@@ -205,4 +228,36 @@ class PersistenceConnectorSpec
...
+        stubForPatchMessageStatus(statusCode)
+
+        whenReady(connector.patchMessageStatus(movementId, messageId, MessageUpdate(MessageStatus.Success)).value) {
+          x =>
+            x.isLeft mustBe true
+
+            statusCode match {
+              case BAD_REQUEST           => x mustBe a[Left[Unexpected, _]]
+              case NOT_FOUND             => x mustBe a[Left[MovementNotFound, _]]
+              case INTERNAL_SERVER_ERROR => x mustBe a[Left[Unexpected, _]]
```
drnaylor-oc, line 257: So this is interesting, I totally get what you've done here but thanks to type erasure on the JVM, this is always going to work because it'll collapse to checking if the object is just a left.

Try this:

```scala
whenReady(connector.patchMessageStatus(movementId, messageId, MessageUpdate(MessageStatus.Success)).value) {
    case Left(error) =>
        statusCode match {
              case BAD_REQUEST           => x mustBe a[Unexpected]
              case NOT_FOUND             => x mustBe a[MovementNotFound]
              case INTERNAL_SERVER_ERROR => x mustBe a[Unexpected]
        }
    case Right(_) => fail("An error was expected")
}
```

That should avoid type erasure issues.

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -1086,4 +1093,68 @@ class MessageControllerSpec
...
+          .thenReturn(EitherT.fromEither(Left(Unexpected())))
+
+        val request = fakeRequestLargeMessage(Json.toJson(sdesResponse), sdesCallback)
+
+        val result = controller().handleSdesResponse()(request)
+
+        status(result) mustBe INTERNAL_SERVER_ERROR
+    }
+
+    "must return BAD_REQUEST for SDES malformed callback" - {
```
drnaylor-oc, line 1152: `-` should be `in`

test/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotificationItemSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import java.time.Instant
+
+class SdesNotificationItemSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPropertyChecks {
+
+  "SuccessfulSubmission" - {
+
+    val instant = Instant.now()
+    "when SdesNotificationItem is serialized, return an appropriate JsObject" - {
+      () =>
```
drnaylor-oc, line 33: "when SdesNotificationItem is serialized, return an appropriate JsObject" in {

and remove the 

```
() =>
```

test/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotificationItemSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+          "checksum"          -> "123",
+          "availableUntil"    -> instant,
+          "dateTime"          -> instant,
+          "properties"        -> Json.obj("name" -> "name", "value" -> "value")
+        )
+        actual mustBe expected
+    }
+
+    "when an appropriate JsObject is deserialized, return SdesNotificationItem" - {
+      () =>
```
drnaylor-oc, line 51: Same here

app/uk/gov/hmrc/transitmovementsrouter/models/sdes/SdesNotification.scala
```diff
@@ -0,0 +1,47 @@
...
+import play.api.libs.json.JsError
+import play.api.libs.json.JsString
+import play.api.libs.json.JsSuccess
+import play.api.libs.json.Json
+import play.api.libs.json.Reads
+import play.api.libs.json.Writes
+
+sealed trait SdesNotification
+
+object SdesNotification {
```
drnaylor-oc, line 28: `SdesNotifcationType`

app/uk/gov/hmrc/transitmovementsrouter/controllers/SdesResponseParser.scala
```diff
@@ -0,0 +1,59 @@
...
+      case SdesNotificationItem(_, _, _, _, _, _, None, _, _) =>
+        logger.info(
+          s"Received a successful response from SDES callback for the following x-conversation-id: ${sdesResponse.conversationId.get.value}"
+        )
+        Future.successful(Right(sdesResponse))
+      case SdesNotificationItem(_, _, _, _, _, _, Some(failureReason), _, _) =>
+        logger.warn(
+          s"Received a failure response from SDES callback for the following x-conversation-id: ${sdesResponse.conversationId.get.value}. Failure reason: $failureReason."
+        )
+        Future.successful(Right(sdesResponse))
```
drnaylor-oc, line 54: I'd log the `SdesNotificationType` too.

it/uk/gov/hmrc/transitmovementsrouter/connectors/PersistenceConnectorSpec.scala
akila-sankar: sure, updated

sonisn1988 (approved):

PR #48, opened 2023-04-11
Report submission status to trader (transit-movements-router)

sonisn1988:


app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -61,7 +62,7 @@ class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2, appC
   private def pushNotificationMessageUpdate(movementId: MovementId, messageId: MessageId): UrlPath =
     UrlPath.parse(s"$baseRoute/traders/movements/${movementId.value}/messages/${messageId.value}")
 
-  override def post(movementId: MovementId, messageId: MessageId, source: Option[Source[ByteString, _]])(implicit
+  override def post(movementId: MovementId, messageId: MessageId, sourceXml: Option[Source[ByteString, _]], sourceJson: Option[JsValue])(implicit
```
drnaylor-oc, line 65: I do not like this at all. We've now got one method that does the following:

* Send an XML payload
* Send a Json payload
* Send no payload

An optional was okay before, but it's not really okay now when we have more complicated situations.

There should now be three methods, one for each scenario. There can be a common private method that they feed in to, no problem, but this is just a messy signature - just have overloaded methods.

drnaylor-oc (changes requested):

olakunlekuye2 (dismissed):

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -61,38 +73,71 @@ class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2, appC
...
+    requestBuilder
+      .execute[Either[UpstreamErrorResponse, HttpResponse]]
+      .map {
+        case Right(_) =>
+          Right(())
+        case Left(error) =>
+          error.statusCode match {
+            case NOT_FOUND             => Left(MovementNotFound(movementId))
+            case INTERNAL_SERVER_ERROR => Left(Unexpected(Some(error.getCause)))
+            case _                     => Left(Unexpected(Some(error.getCause)))
```
drnaylor-oc, line 132: Few things

* `UpstreamErrorResponse` is already a throwable. Just use that.
* Related, I suspect we're doing this elsewhere and you've copied from this, but it's quite possible there will be no cause in the exception, and so you'll end up with `Some(null)`, which we do not want.
* Why have two branches do the same thing, especially as one is the default? I have seen this elsewhere, but that doesn't mean it's right to do.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -177,30 +173,64 @@ class MessagesController @Inject() (
...
+                    MessageStatus.Failed,
+                    Json.toJson(
+                      PresentationError.internalServiceError()
+                    )
+                  )
+                case _ => EitherT.rightT[Future, PresentationError]((): Unit)
+              }
+            } yield persistenceResponse).fold[Result](
+              presentationError => {
+                pushNotificationsConnector.postJSON(movementId, messageId, Json.toJson(presentationError))
```
drnaylor-oc, line 228: If this is a bad request, it's not appropriate to send that to the trader because that's not a them problem, but an us problem.

This **must** be an internal server error.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -177,30 +173,64 @@ class MessagesController @Inject() (
...
-              updateStatus(movementId, messageId, MessageStatus.Success)
-            case SdesNotificationType.FileProcessingFailure =>
-              updateStatus(movementId, messageId, MessageStatus.Failed)
-            case _ => EitherT.rightT[Future, PresentationError]((): Unit)
-          }
-        } yield persistenceResponse).fold[Result](
-          error => Status(error.code.statusCode)(Json.toJson(error)),
-          _ => Ok
-        )
+        parseAndLogSdesResponse(request.body) match {
```
drnaylor-oc, line 195: Add a comment as to why this bit is separate please, just for easier understanding for someone else who comes in (i.e. why is `parseAndLogSdesResponse` separate?)

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -1096,57 +1095,150 @@ class MessageControllerSpec
...
+        mockPersistenceConnector.patchMessageStatus(
+          MovementId(eqTo(movementId.value)),
+          MessageId(eqTo(messageId.value)),
+          eqTo(MessageUpdate(messageStatus))
+        )(any[HeaderCarrier], any[ExecutionContext])
+      )
+        .thenReturn(EitherT.rightT(()))
+      when(
+        mockPushNotificationsConnector
+          .postJSON(MovementId(eqTo(movementId.value)), MessageId(eqTo(messageId.value)), any())(
```
drnaylor-oc, line 1170: We know exactly what `any()` should be, so we should be checking that.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -177,30 +173,64 @@ class MessagesController @Inject() (
...
+            (for {
+              persistenceResponse <- sdesResponse.notification match {
+                case SdesNotificationType.FileProcessed =>
+                  updateAndSendNotification(
+                    movementId,
+                    messageId,
+                    MessageStatus.Success,
+                    Json.toJson(
+                      Json.obj(
+                        "code" -> OK,
```
drnaylor-oc, line 209: What is `OK`? Fairly sure this will return 200, not "OK".

Is "OK" right here, or would something like "SUCCESS" or "PROCESSED" be better?

drnaylor-oc: Actually -- shouldn't this be the Hateoas response that we send from `common-transit-convention-traders`? What do we send on success if a trader sends a small message down the large message route and it succeeds? We need to do the same thing here.

sonisn1988: Here with this endpoint we don't know movementType then how we are going to create href with Hateoas? Also I have discussed this with you and you suggested to use like below so I have used that.
If SDES comes back with a FileProcessingSuccess message, we might need a new message for that, possibly something like:
{
    "code": "ACCEPTED"
    "message": "The message abcdef for movement 0123456 was successfully processed"
}

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
sonisn1988: I just used whatever already there with this class not copied from anywhere. I will change it as suggested above

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
drnaylor-oc: > Also I had discussed this with you and you suggested to use like below so I have used that.

Sure, but now I've reviewed this I've realised we're going to be inconsistent by doing this, and I'd rather not be inconsistent.

> Here with this endpoint we don't know movementType then how we are going to create href with Hateoas?

We control the system, so if we need to figure out a way to do it, then we should.

> If SDES comes back with a FileProcessingSuccess message, we might need a new message for that

...that's what this is though, `FileProcessed` is the success message.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
drnaylor-oc: Yup, fair, didn't show up like that on the side by side diff. May as well change it while we're at it though.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
sonisn1988: I think we can get the movementType by passing movementType with SdesProperties while sending call to Sdes? or another solution I can see by calling transit-movements but not sure if we have endpoint for this on transit-movements without eori number. Please suggest.

drnaylor-oc: Let's take a step back a sec, because I'm thinking about this more and whether what's been done in `common-transit-convention-traders` is correct.

When we send a small message to the large message route, what is the success message we get pushed to us? I see you approved https://github.com/hmrc/common-transit-convention-traders/pull/415, did you look at what the overall result was?

Can you run that, and this, capture the push notification, and paste them here?

drnaylor-oc: I should have checked what the overall result should have been too, of course. I'm concerned that what we've got in the API layer and what we have here are inconsistent, and because we've not been able to test them in regression tests, I'm worried that the wrapping that `transit-movements-push-notifications` adds will make the Hateoas stuff redundant.

Either way, we're not being consistent now and we need to be.

sonisn1988: for router sdes call back(router):

``` {“messageUri”:“/customs/transits/movements/departures/64403215820425ef/messages/6440321586faaf71",“messageBody”:“{\“code\“:\“SUCCESS\“,\“message\“:\“The message 6440321586faaf71 for movement 64403215820425ef was successfully processed\“}”}  ```

for small message via large message route (api):

``` 
{“messageUri”:“/customs/transits/movements/departures/6440332a08662c97/messages/6440332a12db2b6e”,“messageBody”:“{\“_links\“:{\“self\“:{\“href\“:\“/customs/transits/movements/departures/6440332a08662c97/messages/6440332a12db2b6e\“},\“departure\“:{\“href\“:\“/customs/transits/movements/departures/6440332a08662c97\“}}}“}. ```

@drnaylor-oc   Please see the above notification and suggest.

drnaylor-oc: They're both wrong, no-one has really thought about this and so we have a lot of sub-optimal notifications.

Let's think about this and what traders get in a journey for a success.

**SMALL MESSAGE via SMALL MESSAGE API**

Sync response: Hateoas with box ID

**SMALL MESSAGE via LARGE MESSAGE API**

Sync response: Hateoas with box ID + upload information
Push notification: Hateoas that mirrors small message, wrapped in movement/message ID information

**LARGE MESSAGE via LARGE MESSAGE API**

Sync response: Hateoas with box ID + upload information
Push notification: Stringified JSON saying "SUCCESS" wrapped in movement/message ID information

---

Three routes, three different notifications. 

I admit, what I said about returing Hateoas in the large message/large message route doesn't make sense when you take all this in effect, so I apologise about that. However, I don't apologise for catching the inconsistency here -- this is why we should have been testing push notifications earlier because we might have realised this earlier.

I think there are two things that need to happen, and one not need to happen:

* Let's keep what you propose for large messages (SUCCESS etc.) -- that's fine
* We also need to alter the small message/large api route to also return "SUCCESS"
* `transit-movements-push-notifications` need to handle Json payloads differently, so the ultimate payload looks something like:

```json
{
    “messageUri”: “/customs/transits/movements/departures/64403215820425ef/messages/6440321586faaf71",
    "notificationType": "SENT_MESSAGE",
    “response”: {
        "code": "SUCCESS",
        "message": "The message 6440321586faaf71 for movement 64403215820425ef was successfully processed"
    }
} 
```

and for messages going to the trader

```json
{
    “messageUri”: “/customs/transits/movements/departures/64403215820425ef/messages/6440321586faaf71",
    "notificationType": "RECEIVED_MESSAGE",
    "messageBody": "<xml></xml>"
} 
```

By doing that, we make it super easy for traders to differentiate between a notification about a message, and an actual message coming back.

I'll put this into a ticket. For now, just concentrate on this ticket.

sonisn1988: So I hope everything is fine with this ticket just need to alter transit-movements-push-notifications to handle the above cases. right?

drnaylor-oc: > just need to alter transit-movements-push-notifications to handle the above cases. right?

Yes, and a slight change in `common-transit-convention-traders`. I've made tickets for them and passed them to Andy.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -16,6 +16,7 @@
 
 package uk.gov.hmrc.transitmovementsrouter.connectors
 
+import akka.http.scaladsl.model.StatusCodes.Accepted
```
drnaylor-oc, line 19: Remove this -- this is certainly not the `Accepted` you want.

drnaylor-oc (approved):
Before you merge:

* Run **all** regression tests (small AND large) locally and ensure all pass
* Co-ordinate with Kiran and Priya before merging so they are aware and ready for the change to come in.

PR #52, opened 2023-04-18
CTCP-2949: Add ability to log EIS body when a 500 occurs

drnaylor-oc:
We've come across an issue where what we expect to be a 4xx error on ERMIS to be mangled to a 500 by EIS. We want to be able to inspect errors from ERMIS so that we can potentially take action. We may need to turn this off on production, hence it's configurable

### DIFF
--- a/app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
+++ b/app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
@@ -22,6 +22,7 @@ import akka.util.ByteString
 import play.api.Logging
 import play.api.http.HeaderNames
 import play.api.http.MimeTypes
+import play.api.http.Status.INTERNAL_SERVER_ERROR
 import play.api.libs.ws.BodyWritable
 import play.api.libs.ws.SourceBody
 import retry.RetryDetails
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -140,7 +142,15 @@ class EISConnectorImpl(
               logger.info(logMessage + s"Response status: ${result.status}")
               Right(())
             case Left(error) =>
-              logger.warn(logMessage + s"Response status: ${error.statusCode}")
+              val status =
+                if (logBodyOn500 && error.statusCode == INTERNAL_SERVER_ERROR) {
+                  s"""Response status: ${error.statusCode}"
```
derekspalmer, line 147: Does the logged message contain identifying personal info?

derekspalmer (approved):

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
drnaylor-oc: That's what I'm slightly concerned about, I want to see what comes back from QA and make a decision as to whether to turn it off from there (via the config).

PR #56, opened 2023-04-24
CTCP-2571: Allow the router to determine which route to send a message down, not the API

drnaylor-oc:
Part 4a of centralising object-store access.

## What is this PR?

This PR moves EIS/SDES route decision logic to the router and removes access to `common-transit-convention-traders` object store locations (but leaves the `transit-movements-router` one in tact.

## Why?

PlatOps asked us to not access object-store from outside the storing service. This PR implements that (mostly), along with the `common-transit-convention-traders` PR at https://github.com/hmrc/common-transit-convention-traders/pull/418, which must be merged at the same time.

##  What changes are there?

* Whether we submit to SDES or EIS is now decided by the router, not by the API layer, and the returned status code (201 or 202) will tell us how it was submitted so we can act accordingly 
* We no longer accept object store URIs, instead requiring streams
* We will now pass streams to `transit-movements` and not access `object-store` for messages that come in via EIS or SDES.
* Renamed `incoming` and `incomingLargeMessage` to `incomingViaEIS` and `incomingViaUpscan` for easier readability
* Don't log exceptions when XML/Json fails to parse

Regression tests all pass!

## What's next?

Once this is in and working, we'll change the owner in `transit-movements` from `common-transit-convention-traders` to `transit-movements`, and fix `internal-auth-config` to match this.

### DIFF
--- a/app/uk/gov/hmrc/transitmovementsrouter/config/AppConfig.scala
+++ b/app/uk/gov/hmrc/transitmovementsrouter/config/AppConfig.scala
@@ -48,6 +48,7 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
     config.get[String]("microservice.services.object-store.sdes-host")
 
   lazy val logBodyOnEIS500: Boolean = config.get[Boolean]("microservice.services.eis.log-body-on-500")
+  lazy val eisSizeLimit: Long       = config.underlying.getMemorySize("microservice.services.eis.message-size-limit").toBytes
   lazy val logIncoming: Boolean     = config.get[Boolean]("log-incoming-errors")
 
   // SDES configuration
--- a/app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
+++ b/app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
@@ -17,11 +17,13 @@
 package uk.gov.hmrc.transitmovementsrouter.services
 
 import akka.NotUsed
+import akka.stream.Materializer
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import cats.data.EitherT
 import com.google.inject.ImplementedBy
 import play.api.Logging
+import play.mvc.Http.MimeTypes
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.objectstore.client.ObjectSummaryWithMd5
 import uk.gov.hmrc.objectstore.client.Path
@@ -57,6 +59,11 @@ trait ObjectStoreService {
     ec: ExecutionContext
   ): EitherT[Future, ObjectStoreError, ObjectSummaryWithMd5]
 
+  def storeOutgoing(conversationId: ConversationId, stream: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): EitherT[Future, ObjectStoreError, ObjectSummaryWithMd5]
+
   def storeOutgoing(objectStoreResourceLocation: ObjectStoreResourceLocation)(implicit
     hc: HeaderCarrier,
     ec: ExecutionContext
@@ -70,7 +77,9 @@ trait ObjectStoreService {
 }
 
 @Singleton
-class ObjectStoreServiceImpl @Inject() (clock: Clock, appConfig: AppConfig, client: PlayObjectStoreClientEither) extends ObjectStoreService with Logging {
+class ObjectStoreServiceImpl @Inject() (clock: Clock, appConfig: AppConfig, client: PlayObjectStoreClientEither)(implicit mat: Materializer)
+    extends ObjectStoreService
+    with Logging {
   private val dateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss").withZone(ZoneOffset.UTC)
 
   override def storeIncoming(upscanUrl: DownloadUrl, movementId: MovementId, messageId: MessageId)(implicit
@@ -79,7 +88,7 @@ class ObjectStoreServiceImpl @Inject() (clock: Clock, appConfig: AppConfig, clie
   ): EitherT[Future, ObjectStoreError, ObjectSummaryWithMd5] =
     addMessage(
       upscanUrl,
-      Path.File(s"movements/${movementId.value}/${movementId.value}-${messageId.value}-${fileFormat(clock)}.xml"),
+      Path.File(s"movements/${movementId.value}/${movementId.value}-${messageId.value}-$dateFormat.xml"),
       ObjectStoreOwner("common-transit-convention-traders"),
       RetentionPeriod.SevenYears
     )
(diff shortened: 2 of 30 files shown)


app/uk/gov/hmrc/transitmovementsrouter/services/error/UpscanError.scala
```diff
@@ -0,0 +1,24 @@
+/*
```
sonisn1988, line 1: Its better to move this file and router error file to inside ```models/errors```  folder.

it/uk/gov/hmrc/transitmovementsrouter/connectors/UpscanConnectorSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+    with ScalaFutures
+    with ScalaCheckDrivenPropertyChecks {
+
+  implicit val timeout: PatienceConfiguration.Timeout = Timeout(3.seconds)
+
+  "streamFile" - {
+
+    implicit val hc: HeaderCarrier = HeaderCarrier()
+
+    "gets a stream when the file exists" in forAll(Gen.stringOfN(20, Gen.alphaNumChar)) {
```
sonisn1988, line 57: Its better to use ``` alphaNumeric ```  Gen already available inside ``` BaseGenerators ```

app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
```diff
@@ -42,107 +41,41 @@ import javax.inject.Inject
...
-      } yield response).recover {
-        case NonFatal(thr) => Left(ObjectStoreError.UnexpectedError(thr = Some(thr)))
-      }
+      client.putObject[Source[ByteString, _]](
+        Path.File(s"sdes/${conversationId.value.toString}-$dateFormat.xml"),
+        stream,
+        retentionPeriod = RetentionPeriod.OneWeek,
+        contentType = Some(MimeTypes.XML),
+        owner = "transit-movements-router",
+        contentMd5 = None
```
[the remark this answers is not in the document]
drnaylor-oc: It allows for mocking properly, due to the way Scala handles default methods.

app/uk/gov/hmrc/transitmovementsrouter/config/AppConfig.scala
```diff
@@ -48,6 +48,7 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
     config.get[String]("microservice.services.object-store.sdes-host")
 
   lazy val logBodyOnEIS500: Boolean = config.get[Boolean]("microservice.services.eis.log-body-on-500")
+  lazy val eisSizeLimit: Long       = config.underlying.getMemorySize("microservice.services.eis.message-size-limit").toBytes
```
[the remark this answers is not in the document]
drnaylor-oc: No, this is absolutely correct and intended. Have a look at the configuration file -- it's for parsing values of that format.

More readable -- that's far better.

it/uk/gov/hmrc/transitmovementsrouter/connectors/UpscanConnectorSpec.scala
drnaylor-oc: Why is it better to use that, rather than an inbuilt?

sonisn1988: if there is not benefit then what is the purpose of creating below one inside BaseGenerators?
  def alphaNumeric(length: Int): Gen[String] =
    for {
      listOfChar <- listOfN(length, alphaNumChar)
    } yield listOfChar.mkString

it/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesControllerIntegrationSpec.scala
```diff
@@ -130,144 +161,251 @@ class MessagesControllerIntegrationSpec
...
+  private val objectStoreService = mock[PlayObjectStoreClientEither]
+  when(
+    objectStoreService.putObject(
+      any[Path.File],
+      any[Source[ByteString, _]],
+      eqTo(RetentionPeriod.OneWeek),
+      eqTo(Some(MimeTypes.XML)),
+      eqTo(None),
+      eqTo("transit-movements-router")
+    )(any(), any[HeaderCarrier])
```
sonisn1988, line 180: Please use any[ExecutionContext] here instead of any()  to keep consistent with HeaderCarrier

it/uk/gov/hmrc/transitmovementsrouter/connectors/UpscanConnectorSpec.scala
drnaylor-oc: Because someone didn't realise that they could do what I did.

> and also I think keeping generator at one place make more sense.

We should stick with the base library one then, and not duplicate it with our own.

it/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesControllerIntegrationSpec.scala
```diff
@@ -130,144 +161,251 @@ class MessagesControllerIntegrationSpec
...
+        val conversationId          = ConversationId(UUID.randomUUID())
+        val (movementId, messageId) = conversationId.toMovementAndMessageId
+
+        val time      = OffsetDateTime.of(2023, 2, 14, 15, 55, 28, 0, ZoneOffset.UTC)
+        val formatted = DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss z", Locale.ENGLISH).withZone(ZoneOffset.UTC).format(time)
+
+        // should not happen
+        server.stubFor(
+          post(
+            urlEqualTo("/xi")
```
sonisn1988, line 267: same here please add withRequestBody

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -207,22 +226,53 @@ class MessageControllerSpec
...
+      arbitrary[ObjectSummaryWithMd5]
+    ) {
+      (eori, movementType, movementId, messageId, summary) =>
+        val expectedConversationId = ConversationId(movementId, messageId)
+
+        when(config.eisSizeLimit).thenReturn(-1)
+
+        when(mockMessageTypeExtractor.extractFromHeaders(any())).thenReturn(EitherT.rightT[Future, MessageTypeExtractionError](MessageType.DeclarationData))
+
+        when(mockCustomOfficeExtractorService.extractCustomOffice(any(), any()))
```
sonisn1988, line 253: better to pass the requestMessageType instead of any

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -207,22 +226,53 @@ class MessageControllerSpec
...
+
+        status(result) mustBe ACCEPTED
+
+        verify(mockRoutingService, times(0)).submitMessage(
+          any[String].asInstanceOf[MovementType],
+          any[String].asInstanceOf[MovementId],
+          any[String].asInstanceOf[MessageId],
+          any[Source[ByteString, _]],
+          any[String].asInstanceOf[CustomsOffice]
+        )(any[HeaderCarrier], any[ExecutionContext])
```
sonisn1988, line 275: Its also good to verify sdes service call should be called once here.

test/uk/gov/hmrc/transitmovementsrouter/generators/TestModelGenerators.scala
```diff
@@ -51,14 +51,35 @@ trait TestModelGenerators extends BaseGenerators {
...
+    Arbitrary {
+      for {
+        country <- Gen.oneOf("GB", "XI")
+        code    <- Gen.stringOfN(10, Gen.numChar)
+      } yield EoriNumber(s"$country$code")
+    }
+
+  implicit lazy val arbitraryMovementType: Arbitrary[MovementType] =
+    Arbitrary {
+      Gen.oneOf(MovementType("departure"), MovementType("arrival"))
```
sonisn1988, line 64: This should be either ```departures ``` or ``` arrivals ```

it/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesControllerIntegrationSpec.scala
drnaylor-oc: That said, the comment is wrong, so that needs fixing

it/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesControllerIntegrationSpec.scala
drnaylor-oc: It's not an `ExecutionContext`, but an `ObjectStoreContentWrite[F, CONTENT, REQ_BODY]`, where all three type parameters are actually generics. Due to kind projection shenanigans (I think) there's no way I can put the type in here, so I won't.

sonisn1988 (approved):

PR #57, opened 2023-04-27
CTCP-3004: Allow EIS stub to act as a proxy for selectively sending messages to EIS, or remaining in stub mode

liandra-giacomine:


app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -103,7 +103,10 @@ class EISConnectorImpl(
       onFailure
     ) {
       // blank-ish carrier so that we control what we're sending to EIS, and let the carrier/platform do the rest
-      implicit val headerCarrier: HeaderCarrier = HeaderCarrier(requestChain = hc.requestChain)
+      implicit val headerCarrier: HeaderCarrier =
+        if (eisInstanceConfig.forwardClientId)
+          HeaderCarrier(requestChain = hc.requestChain, extraHeaders = hc.headers(Seq("X-Client-Id")))
```
drnaylor-oc, line 108: Can we get a test the checks this works and the request in question actually sends the client ID? I've had issues where this doesn't happen because we the `HeaderCarrier` is a bit of a mess, so I'd rather have the confidence with this backed up in a test.

drnaylor-oc (approved):
Will merge once E2E testing is done for the day

drnaylor-oc (approved):

PR #65, opened 2023-06-21
CTCP-3298 restrict duplicate LRN from EIS

akila-sankar:


app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -108,11 +111,12 @@ class EISConnectorImpl(
...
-      val correlationId  = UUID.randomUUID().toString
-      val accept         = MimeTypes.XML
-      val conversationId = ConversationId(movementId, MessageId("0000000000000000")) // ConversationId(movementId, messageId)
-      val date           = HTTP_DATE_FORMATTER.format(OffsetDateTime.now())
+      val requestId       = hc.requestId.map(_.value)
+      val correlationId   = UUID.randomUUID().toString
+      val accept          = MimeTypes.XML
+      val conversationId  = ConversationId(movementId, MessageId("0000000000000000")) // ConversationId(movementId, messageId)
+      val date            = HTTP_DATE_FORMATTER.format(OffsetDateTime.now())
+      val lrnRegexPattern = "The supplied LRN: ([a-zA-Z0-9]+) has already been used by submitter: ([a-zA-Z0-9]+)".r
```
drnaylor-oc, line 119: This variable should be a lazy val on the class level as this is a static regex that won't change. That way we calculate it once, rather than for each request.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -157,7 +161,30 @@ class EISConnectorImpl(
...
+              error.statusCode match {
+                case ErrorCode.Forbidden.statusCode | ErrorCode.InternalServerError.statusCode =>
+                  val lrnMatch = lrnRegexPattern.findFirstMatchIn(error.message)
+                  lrnMatch match {
+                    case Some(message) =>
+                      val lrn: LocalReferenceNumber =
+                        Some(message)
+                          .map(_.group(1))
+                          .map(LocalReferenceNumber.apply)
+                          .get
```
drnaylor-oc, line 174: Why not just `LocalReferenceNumber(message.group(1))`?

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -157,7 +161,30 @@ class EISConnectorImpl(
...
+                      Left(RoutingError.DuplicateLRNError(s"LRN ${lrn.value} has previously been used and cannot be reused", ErrorCode.Conflict, lrn))
+                    case None =>
+                      Left(
+                        RoutingError.Upstream(
+                          UpstreamErrorResponse(s"Request Error: Routing to $code returned status code ${error.statusCode}", error.statusCode)
+                        )
+                      )
+                  }
```
drnaylor-oc, line 182: You could reduce this down to 

```suggestion
                  message match {
                    case lrnRegexPattern(lrn) =>
                      Left(RoutingError.DuplicateLRNError(s"LRN $lrn has previously been used and cannot be reused", ErrorCode.Conflict, LocalReferenceNumber(lrn)))
                    case None =>
                      Left(
                        RoutingError.Upstream(
                          UpstreamErrorResponse(s"Request Error: Routing to $code returned status code ${error.statusCode}", error.statusCode)
                        )
                      )
                  }
```

See https://docs.scala-lang.org/tour/regular-expression-patterns.html -- it's quite a cool (in my opinion!) use of pattern matching in Scala. `case lrnRegexPattern(lrn)` matches the string, and extracts the first group as `lrn`.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
akila-sankar: Yes agree, updated it.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
akila-sankar: Yes @drnaylor-oc totally agree, this is how I first implemented. But below is the response from eis-stub for the _error.message_ attribute and it did not extracts the first group as `lrn`. 

`POST of 'http://localhost:9476/transit-movements-eis-stub/movements/messages/gb' returned 403. Response body: '{"code":"FORBIDDEN","message":"Error in request: The supplied LRN: LRN12345678 has already been used by submitter: messageSender"}'`

So had to tweak it accordingly. Welcome for any other suggestions.

drnaylor-oc: Ah, yeah, because of that extra stuff the error puts around it.

I just wrote a quick scala-cli script that does work:

```scala
//> using scala "2.13"

val pattern = ".*The supplied LRN: ([a-zA-Z0-9]+) has already been used by submitter: ([a-zA-Z0-9]+).*".r

val text = """POST of 'http://localhost:9476/transit-movements-eis-stub/movements/messages/gb' returned 403. Response body: '{"code":"FORBIDDEN","message":"Error in request: The supplied LRN: LRN12345678 has already been used by submitter: messageSender"}'"""

text match {
    case pattern(t, _) => println(s"This worked: LRN $t")
    case _ => println("This didn't work")
}
```

Two things:

* We need to surround the regex with `.*` -- so it'll do a full match on the entire string
* The unapply needs two parameters as there are two groups in your regex (which I didn't notice!)

akila-sankar: Ah yes I could have updated the regex, didn't realize it. Thank you @drnaylor-oc will update it now.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -89,8 +93,9 @@ class EISConnectorImpl(
 
   def onFailure(response: Either[RoutingError, Unit], retryDetails: RetryDetails): Future[Unit] =
     response.left.toOption.get match {
-      case RoutingError.Upstream(upstreamErrorResponse) => attemptRetry(s"with status code ${upstreamErrorResponse.statusCode}", retryDetails)
-      case RoutingError.Unexpected(message, _)          => attemptRetry(s"with error $message", retryDetails)
+      case RoutingError.Upstream(upstreamErrorResponse)       => attemptRetry(s"with status code ${upstreamErrorResponse.statusCode}", retryDetails)
+      case RoutingError.DuplicateLRNError(message, code, lrn) => attemptRetry(s"with status code $code and error $message", retryDetails)
```
drnaylor-oc, line 97: We definitely do not want to retry on a duplicate LRN error, because we know that will never succeed.

drnaylor-oc (commented):
A duplicate LRN error from EIS is an error that we can't do anything about, and retrying won't help with. It shouldn't trigger the circuit breaker either -- so two things need to happen:

* `shouldCauseCircuitBreakerStrike` should return `true` if the `RoutingError` is a `DuplicateLRNError`, stopping the circuit breaker from potentially tripping erroneously.
* Check this, but I think the lambda in `retryingOnFailures` needs to return `Future(true)` if the error is a `DuplicateLRNError` so we don't retry sending the same LRN again.

I'd also like there to be tests around these if possible too - they might be complex to set up but this is a slightly complicated condition and tests will enforce it, ensuring others who touch this code will see that this is intended.

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
akila-sankar: Yes you are right, updated it and added test as well to cover this scenario

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -91,11 +91,7 @@ class EISConnectorImpl(
     s"${HTTP_DATE_FORMATTER.format(OffsetDateTime.now(clock.withZone(ZoneOffset.UTC)))} UTC"
 
   def shouldCauseCircuitBreakerStrike(result: Try[Either[RoutingError, Unit]]): Boolean =
-    result match {
-      case Success(Left(DuplicateLRNError(_, _, _))) => false // Not to open circuit breaker for DuplicateLRN error
-      case Success(_)                                => true
-      case _                                         => false
-    }
+    result.map(_.isLeft).getOrElse(true)
```
drnaylor-oc: We don't want a circuit breaker strike for a duplicate LRN error -- why go back to this?

akila-sankar: Yes @drnaylor-oc this is controlled in retryingOnFailures method & circuit breaker won't be triggered.

akila-sankar: @drnaylor-oc Thank you for the detailed explanation. I understood it bit slightly different as when I tested it didn't do the retry & didn't go inside the circuit breaker and now it makes more clear now. Updated the code and will be pushing the changes after tests run.

derekspalmer (approved):

app/uk/gov/hmrc/transitmovementsrouter/connectors/EISConnector.scala
```diff
@@ -99,7 +109,11 @@ class EISConnectorImpl(
   override def post(movementId: MovementId, messageId: MessageId, body: Source[ByteString, _], hc: HeaderCarrier): Future[Either[RoutingError, Unit]] =
     retryingOnFailures(
       retries.createRetryPolicy(eisInstanceConfig.retryConfig),
-      (t: Either[RoutingError, Unit]) => Future.successful(t.isRight),
+      (t: Either[RoutingError, Unit]) =>
+        t match {
+          case Left(RoutingError.DuplicateLRNError(_, _, _)) => Future.successful(t.isLeft) //Not to want retry for DuplicateLRN error
+          case _                                             => Future.successful(t.isRight)
```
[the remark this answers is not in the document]
drnaylor-oc: That's okay, this instructs the retry system to retry if it's not a right, which is what we want after the more specific match prior.

app/uk/gov/hmrc/transitmovementsrouter/controllers/errors/ConvertError.scala
```diff
@@ -58,9 +58,10 @@ trait ConvertError {
   implicit val routingErrorConverter = new Converter[RoutingError] {
 
     def convert(routingError: RoutingError): PresentationError = routingError match {
-      case Upstream(upstreamErrorResponse) => PresentationError.internalServiceError(cause = Some(upstreamErrorResponse.getCause))
-      case Unexpected(_, cause)            => PresentationError.internalServiceError(cause = cause)
-      case BadDateTime(element, ex)        => PresentationError.badRequestError(s"Could not parse datetime for $element: ${ex.getMessage}")
+      case Upstream(upstreamErrorResponse)       => PresentationError.internalServiceError(cause = Some(upstreamErrorResponse.getCause))
+      case Unexpected(_, cause)                  => PresentationError.internalServiceError(cause = cause)
+      case BadDateTime(element, ex)              => PresentationError.badRequestError(s"Could not parse datetime for $element: ${ex.getMessage}")
+      case DuplicateLRNError(message, code, lrn) => PresentationError.duplicateLRNError(message, lrn)
```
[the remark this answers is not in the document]
akila-sankar: Sure, will update

derekspalmer (approved):

PR #66, opened 2023-06-26
CTCP 3210: Integrate NCTS monitoring tool

drnaylor-oc:
This completely follows the P4 spec and there will be a couple of things that will need to be done in the tool itself.

Note that it accepts a content type of `text/plain` only(!), so we need to force that, even though we send Json (and do for P4)...

### DIFF
--- a/app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
+++ b/app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
@@ -73,7 +73,11 @@ class ObjectStoreServiceImpl @Inject() (clock: Clock, appConfig: AppConfig, clie
         contentMd5 = None
       )
     }.leftMap {
-      case NonFatal(thr) => ObjectStoreError.UnexpectedError(thr = Some(thr))
+      error =>
+        // avoids exhaustivity warnings
+        (error: @unchecked) match {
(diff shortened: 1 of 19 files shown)


app/uk/gov/hmrc/transitmovementsrouter/services/ObjectStoreService.scala
```diff
@@ -73,7 +73,11 @@ class ObjectStoreServiceImpl @Inject() (clock: Clock, appConfig: AppConfig, clie
         contentMd5 = None
       )
     }.leftMap {
-      case NonFatal(thr) => ObjectStoreError.UnexpectedError(thr = Some(thr))
+      error =>
+        // avoids exhaustivity warnings
+        (error: @unchecked) match {
```
derekspalmer, line 78: :+1:

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -716,6 +757,13 @@ class MessageControllerSpec
 
         verifyNoInteractions(mockPushNotificationsConnector)
         verifyNoInteractions(mockInternalAuthActionProvider)
+
+        // It was our fault, not other bits, so we should still report availability
```
derekspalmer, line 761: :+1:

derekspalmer (approved):
LGTM.

PR #68, opened 2023-07-10
CTCP-3517: Add message type header to PPNS calls, update URLs

drnaylor-oc:


app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -72,83 +72,73 @@ class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2)(impl
...
+  override def postMessageReceived(movementId: MovementId, messageId: MessageId, messageType: MessageType, sourceXml: Source[ByteString, _])(implicit
     hc: HeaderCarrier,
     ec: ExecutionContext
   ): EitherT[Future, PushNotificationError, Unit] =
     EitherT {
       if (!appConfig.pushNotificationsEnabled) {
         Future.successful(Right(()))
       } else {
-        val request = createRequest(movementId, messageId)
+        val request = createRequest(movementId, messageId, "messageReceived")
```
sonisn1988, line 86: I think we can create class just like ``` MovementType ``` instead of passing string directly.

drnaylor-oc: if this constant is only being used here, why?

sonisn1988: Hardcoding it at multiple places might can create some mistake for anyone else if does not get caught in tests. It's my thinking and suggestion. If does not seems good then I am fine with that

drnaylor-oc: Sure, it's being used twice in actual fact, so I've put it into a constant in the class.

sonisn1988 (approved):

derekspalmer (approved):

PR #70, opened 2023-10-11
[CTCP-4066] added audit connector and audit service

prashanthmamidi:


test/uk/gov/hmrc/transitmovementsrouter/utils/TestCommonGenerators.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.scalacheck.Arbitrary
+import org.scalacheck.Gen
+import uk.gov.hmrc.transitmovementsrouter.models.AuditType
+import uk.gov.hmrc.transitmovementsrouter.models.EoriNumber
+import uk.gov.hmrc.transitmovementsrouter.models.MessageId
+import uk.gov.hmrc.transitmovementsrouter.models.MessageType
+import uk.gov.hmrc.transitmovementsrouter.models.MovementId
+import uk.gov.hmrc.transitmovementsrouter.models.MovementType
+
+trait TestCommonGenerators {
```
drnaylor-oc, line 28: Some of these are also in `TestModelGenerators`. Rather than redefine them like this, use what exists already.

app/uk/gov/hmrc/transitmovementsrouter/models/AuditType.scala
```diff
@@ -0,0 +1,86 @@
...
+  final case object RequestOnNonArrivedMovement                        extends AuditType("RequestOnNonArrivedMovement")
+  final case object InformationAboutNonArrivedMovement                 extends AuditType("InformationAboutNonArrivedMovement")
+  final case object PresentationNotificationForThePreLodgedDeclaration extends AuditType("PresentationNotificationForThePreLodgedDeclaration")
+  final case object FunctionalNack                                     extends AuditType("FunctionalNack")
+  final case object PositiveAcknowledge                                extends AuditType("PositiveAcknowledge")
+  final case object ArrivalNotification                                extends AuditType("ArrivalNotification")
+  final case object ForwardedIncidentNotificationToED                  extends AuditType("ForwardedIncidentNotificationToED")
+  final case object LargeMessageSubmissionRequested                    extends AuditType("LargeMessageSubmissionRequested")
+  final case object TraderFailedUploadEvent                            extends AuditType("TraderFailedUploadEvent")
```
drnaylor-oc, line 50: These aren't appropriate here, so we shouldn't include them.

app/uk/gov/hmrc/transitmovementsrouter/models/AuditType.scala
```diff
@@ -0,0 +1,86 @@
...
+  final case object RejectionFromOfficeOfDestination                   extends AuditType("RejectionFromOfficeOfDestination")
+  final case object ControlDecisionNotification                        extends AuditType("ControlDecisionNotification")
+  final case object RequestOnNonArrivedMovement                        extends AuditType("RequestOnNonArrivedMovement")
+  final case object InformationAboutNonArrivedMovement                 extends AuditType("InformationAboutNonArrivedMovement")
+  final case object PresentationNotificationForThePreLodgedDeclaration extends AuditType("PresentationNotificationForThePreLodgedDeclaration")
+  final case object FunctionalNack                                     extends AuditType("FunctionalNack")
+  final case object PositiveAcknowledge                                extends AuditType("PositiveAcknowledge")
+  final case object ArrivalNotification                                extends AuditType("ArrivalNotification")
+  final case object ForwardedIncidentNotificationToED                  extends AuditType("ForwardedIncidentNotificationToED")
```
drnaylor-oc, line 48: Only some of these will be important for this service, so we shouldn't include those that are actually for the API layer. Equally, if the API layer has some of these that should only be here on the router, we should make sure we remove them from there.

app/uk/gov/hmrc/transitmovementsrouter/connectors/AuditingConnector.scala
```diff
@@ -0,0 +1,127 @@
...
+import uk.gov.hmrc.transitmovementsrouter.models.MovementType
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
+@ImplementedBy(classOf[AuditingConnectorImpl])
+trait AuditingConnector {
+
+  def post(
```
sonisn1988, line 50: @drnaylor-oc / @prashanthmamidi are we not going to include statusType audit event here? Con't find them but it is requited I think?

prashanthmamidi: @sonisn1988 yes it's needed, will add this.  I think when I copied this file from CTC traders  service this audit status Type changes are not available at that time.

app/uk/gov/hmrc/transitmovementsrouter/models/AuditType.scala
derekspalmer: Apart from going through the process flows, is this info to hand in any of our analysis docs?

drnaylor-oc: Yes, that big Confluence doc that I wrote to pull everything together and got everyone to review.

prashanthmamidi: updated the auditType list

app/uk/gov/hmrc/transitmovementsrouter/models/AuditType.scala
derekspalmer, line 1: :+1:

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
```diff
@@ -16,6 +16,12 @@
...
 
+import play.api.libs.json.JsError
+import play.api.libs.json.JsString
+import play.api.libs.json.JsSuccess
+import play.api.libs.json.Reads
+import play.api.libs.json.Writes
+
 sealed trait MessageType extends Product with Serializable {
   def code: String
   def rootNode: String
```
derekspalmer, line 27: Don't we need associated auditType and movementType ?

app/uk/gov/hmrc/transitmovementsrouter/models/MovementType.scala
```diff
@@ -16,4 +16,11 @@
 
 package uk.gov.hmrc.transitmovementsrouter.models
 
+import play.api.libs.json.Format
+import play.api.libs.json.Json
+
 case class MovementType(value: String) extends AnyVal
```
derekspalmer, line 22: 2 movement objects:   Departure and Arrival?

derekspalmer (commented):
Just a couple of queries.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
prashanthmamidi: I think you mean MessageTypes -  this is an existing class, not sure we can update as part of this ticket

app/uk/gov/hmrc/transitmovementsrouter/models/MovementType.scala
prashanthmamidi: again, it's an existing class - not sure will add any value now for this as we end up changing loads of files if we change this.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
drnaylor-oc: I'm in two minds about this, because really, the audit types themselves weren't part of this ticket, just creating the connector and service was. Another ticket was about firing the message type audit events and I'd have expected this type association to be done then -- but I'd have also expected the `AuditType`s to have been added in another ticket too.

Let's make this ticket about the service/connector only, and make sure it matches what the API service/connectors look like. The associations can be done in the ticket for firing message type messages.

derekspalmer (approved):

PR #73, opened 2023-10-17
Ctcp 4069

derekspalmer:


app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -135,19 +138,36 @@ class MessagesController @Inject() (
...
     authenticateEISToken.stream(transformer = eisMessageTransformers.unwrap) {
-      implicit request => _ =>
+      implicit request => size =>
         import MessagesController.PresentationEitherTHelper
 
         val (movementId, triggerId) = ids.toMovementAndMessageId
-
         (for {
           messageType <- messageTypeExtractor.extract(request.headers, request.body).asPresentationWithMessageType(None)
+          auditMsg    <- extractAuditMessageType(messageType).asPresentationWithMessageType(None)
```
sonisn1988, line 155: I don't think so we need to send auditType with ```None```

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
```diff
@@ -22,14 +22,21 @@ import play.api.libs.json.JsSuccess
 import play.api.libs.json.Reads
 import play.api.libs.json.Writes
 
+//TODO: need to take a look at this with respect to whether auditType and movementType needed at this level.
 sealed trait MessageType extends Product with Serializable {
   def code: String
   def rootNode: String
+  def auditType: Option[AuditType] = None
```
sonisn1988, line 29: Better to use this auditType at ```DepartureResponseMessageType``` and ``` ArrivalResponseMessageType``` . As we always need an AuduitType to audit the event so I prefer to make it mandatory.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
```diff
@@ -22,14 +22,21 @@ import play.api.libs.json.JsSuccess
 import play.api.libs.json.Reads
 import play.api.libs.json.Writes
 
+//TODO: need to take a look at this with respect to whether auditType and movementType needed at this level.
```
sonisn1988, line 25: Please remove this TODO post changes in this file.

app/uk/gov/hmrc/transitmovementsrouter/models/MovementType.scala
```diff
@@ -23,4 +23,8 @@ case class MovementType(value: String) extends AnyVal
 
 object MovementType {
   implicit lazy val movementTypeFormat: Format[MovementType] = Json.valueFormat[MovementType]
+
+  val Departure = MovementType("departure")
```
sonisn1988, line 27: Why we need to create this two additional variables?  I think we can directly pass string and use case class constructor?

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
```diff
@@ -165,6 +167,7 @@ class MessageControllerSpec
       mockSDESService,
       mockInternalAuthActionProvider,
       mockStatusMonitoringService,
+      mockAuditingService,
```
sonisn1988, line 170: can not see any test cases to verify the auditing call with messagetType payload. Please add them.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -135,19 +138,36 @@ class MessagesController @Inject() (
...
   def incomingViaEIS(ids: ConversationId): Action[Source[ByteString, _]] =
     authenticateEISToken.stream(transformer = eisMessageTransformers.unwrap) {
-      implicit request => _ =>
+      implicit request => size =>
         import MessagesController.PresentationEitherTHelper
 
         val (movementId, triggerId) = ids.toMovementAndMessageId
-
         (for {
           messageType <- messageTypeExtractor.extract(request.headers, request.body).asPresentationWithMessageType(None)
```
sonisn1988, line 154: Here with this call we are not validating about request and response type messages. So here first we need to make sure it is response type message then we need to calculate the audittype. As for request type message we are already doing auditing at api side so need to do auditing for request type message.

sonisn1988 (changes requested):
Please add the necessary test cases and changes as suggested.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
drnaylor-oc: That's not what this doing -- it's just for error handling.

That said, this should be `asPresentationWithMessageType(Some(messageType))`.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
drnaylor-oc: If EIS sends us a request message, we have bigger problems. If we want to do that validation, that's a tech debt ticket, but I don't really care either way.

drnaylor-oc: ...though that's done in the extractAuditMessageType function anyway?

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -135,19 +138,36 @@ class MessagesController @Inject() (
     }
   }
 
+  private def extractAuditMessageType(messageType: MessageType): EitherT[Future, MessageTypeExtractionError, AuditType] = {
+    val auditType: EitherT[Future, MessageTypeExtractionError, AuditType] =
+      EitherT.fromOption[Future](messageType.auditType, MessageTypeExtractionError.InvalidMessageType(s"$messageType is not a ResponseMessageType"))
+    auditType
```
drnaylor-oc, line 144: Get rid of the local variable, this can be a one liner -- though really, as this is in the presentation layer, instead of returning a `MessageTypeExtractionError`, just return a `PresentationError` directly.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
drnaylor-oc: I can see the value in keeping it like this, considering how the audit type is extracted -- it's main draw for me is that it avoids a type match later, which can be expensive (reflection!) compared to a simple dynamic dispatch.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
derekspalmer: It's not sending the audit type.

I'll change to directly return the presentation error.

app/uk/gov/hmrc/transitmovementsrouter/models/MessageTypes.scala
derekspalmer: The reason I put it at this level because of the the way in which the audit type is extracted, it's cheaper in terms of resources at this level, and it stops it becoming a ticket about tech debt rather than auditing.
That's why the TODO was there: a reminder as to whether we should to take a look at how our message types are structured.

app/uk/gov/hmrc/transitmovementsrouter/models/MovementType.scala
derekspalmer: First we shouldn't pass Strings, second, I would have preferred a Departure and Arrival object, but then it becomes a refactoring ticket.

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
derekspalmer: We should NOT get a request message, otherwise we are in trouble! That's why auditType is set to None as default for request - we then get protection from the call to extractAuditMessage.

test/uk/gov/hmrc/transitmovementsrouter/controllers/MessageControllerSpec.scala
derekspalmer: Agree. Will add verification of audit service calls.

sonisn1988 (approved):

PR #74, opened 2023-10-17
Ctcp 4181 Extra P4 audit events for failed scenario and CTCP-4180 Create audit event for success event

sonisn1988:


derekspalmer (commented):
I'm a bit puzzled by this ticket as it totally replaces the work just completed on CTCP-4069!

sonisn1988:
> CTCP-4069



> I'm a bit puzzled by this ticket as it totally replaces the work just completed on CTCP-4069!
This ticket is to add success status type event post data is successful persisted in db. CTCP-4069 to audit message type event post data is persisted in db.and with this both type are events are there. Where are you puzzled it is replaced with CTCP-4069? Also with this ticket I have added one failure status type event in case of movement does not exist in db.

PR #81, opened 2024-06-04
fix for java.nio.file.NoSuchFileException: /tmp/playtemp1442994242346491910/8887008831816427304

sonisn1988:


app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -122,35 +123,125 @@ class MessagesController @Inject() (
         _               <- sdesService.send(movementId, messageId, objectStoreFile).asPresentation
       } yield Accepted
 
-    internalAuth(predicate).streamWithSize {
-      implicit request => size =>
+    internalAuth(predicate).async(streamFromMemory) {
+
+      implicit request =>
         (for {
+          source             <- reUsableOutgoingSource(request)
```
derekspalmer, line 130: Rather than source have sourceCopies

app/uk/gov/hmrc/transitmovementsrouter/controllers/MessagesController.scala
```diff
@@ -122,35 +123,125 @@ class MessagesController @Inject() (
...
+        }
+    )
+
+  // Function to create a new source from the materialized sequence
+  private def createReusableSource(seq: Seq[ByteString]): Source[ByteString, _] = Source(seq.toList)
+
+  // Materialize the source and create multiple new sources with error handling
+  private def reUsableIncomingSource(request: Request[Source[ByteString, _]]): EitherT[Future, PresentationError, List[Source[ByteString, _]]] = for {
+    byteStringSeq <- materializeIncomingSource(request.body)
+  } yield List.fill(4)(createReusableSource(byteStringSeq))
```
derekspalmer, line 198: Rather than a magic number 4 and so we know why, can we declare something like
val numberOfSourceCopiesRequired = 4

List.fill(numberOfSourceCopiesRequired)

derekspalmer (approved):
I can see what you have done and why it works but I'm uncomfortable with making 4 copies of the stream  - we need to make sure performance tests are executed and checked. Also,  need to ensure no problems in External Test.

PR #99, opened 2025-02-17
CTCP-6273: Add condition to determine whether to send push notification

DamienButterworth:


Arei1997 (approved):

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -82,15 +82,20 @@ class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2)(impl
...
     hc: HeaderCarrier,
     ec: ExecutionContext
   ): EitherT[Future, PushNotificationError, Unit] =
     EitherT {
-      if (!appConfig.pushNotificationsEnabled) {
+      if (!appConfig.pushNotificationsEnabled || !persistenceResponse.shouldSendNotification) {
         Future.successful(Right(()))
       } else {
-        val request = createRequest(movementId, messageId, messageReceivedPath)
+        val request = createRequest(movementId, persistenceResponse.messageId, messageReceivedPath)
```
sonisn1988, line 98: we need to call this part based on persistenceResponse service boolean value. That part is missing and need to fix it.

app/uk/gov/hmrc/transitmovementsrouter/connectors/PushNotificationsConnector.scala
```diff
@@ -82,15 +82,20 @@ class PushNotificationsConnectorImpl @Inject() (httpClientV2: HttpClientV2)(impl
...
+    persistenceResponse: PersistenceResponse,
+    messageType: MessageType,
+    sourceXml: Source[ByteString, ?]
+  )(implicit
     hc: HeaderCarrier,
     ec: ExecutionContext
   ): EitherT[Future, PushNotificationError, Unit] =
     EitherT {
-      if (!appConfig.pushNotificationsEnabled) {
+      if (!appConfig.pushNotificationsEnabled || !persistenceResponse.shouldSendNotification) {
```
sonisn1988, line 95: This part only handle if the ppns notification is disabled then only just return Unit. Nothing else. So from here we can remove or condition. That does not matter at all.

PR #121, opened 2026-08-13
Investigate Optional Header for Router Redirection

sandeep-k-soni:


app/uk/gov/hmrc/transitmovementsrouter/services/RoutingService.scala
```diff
@@ -57,16 +58,19 @@ class RoutingServiceImpl @Inject() (
...
     )
 
-  private def eisConnectorSelector(customsOffice: CustomsOffice, versionHeader: APIVersionHeader): EISConnector =
-    (customsOffice.isGB, versionHeader) match {
-      case (true, APIVersionHeader.v3_0)  => messageConnectorProvider.gbV3_0
-      case (false, APIVersionHeader.v3_0) => messageConnectorProvider.xiV3_0
+  private def eisConnectorSelector(customsOffice: CustomsOffice, versionHeader: APIVersionHeader, optionalHeader: Option[String]): EISConnector =
+    (customsOffice.isGB, versionHeader, optionalHeader) match {
+      case (true, APIVersionHeader.v3_0, Some(optionalHeader))  => logger.info("routing to optional header GB route"); messageConnectorProvider.gbV3_1
+      case (false, APIVersionHeader.v3_0, Some(optionalHeader)) => logger.info("routing to optional header XI route"); messageConnectorProvider.xiV3_1
```
KiraNamida, line 72: This appears more like we using the presence of the optional header to redirect rather than the version within the header but I can see that would be more work here as it points to a predefined connector.
Might be something to look at more post Spike.

KiraNamida (approved):

